PR #256, opened 2026-01-15
fix(ui): remove max-width constraint from agent edit page

chaliy:
## What
Remove the `max-w-4xl` constraint from the Agent edit page to match the Create Agent page width.

## Why
The Edit Agent page had a narrower max-width than the Create Agent page, causing inconsistent UI experience.

## How
Removed the `max-w-4xl` class from the container div in the edit page.

## Risk
- Low
- Visual change only, no functional impact

## Checklist
- [x] Tests added or updated


chaliy:
## 📸 UI Screenshot

**Create Agent page - shows full width layout (same as Edit page after fix)**

Create Agent page - shows full width layout (same as Edit page after fix)

PR #322, opened 2026-01-20
feat(ui): move info icon to footer row in chat messages

chaliy:
## What
Move MessageInfoIcon from inline position to a dedicated footer row below message content in chat messages.

## Why
The previous inline positioning caused ugly, inconsistent layouts - the info icon would appear at different positions depending on message length.

## How
- Added a footer row (`flex justify-end gap-2 pt-1`) below message content for both user and agent messages
- This provides consistent positioning and room for future additions (timestamps, token counts, etc.)

## Risk
- Low
- Visual-only change, no logic changes

## Checklist
- [x] UI build passes
- [x] Dev showcase updated to reflect new pattern


chaliy:
## 📸 UI Screenshot

**Chat page with footer row layout for info icons - showing both user and agent messages**

Chat page with footer row layout for info icons - showing both user and agent messages

PR #500, opened 2026-02-16
feat(sessions): add per-user session pinning

chaliy:
## What
Add ability to pin sessions per user on the session page. Pinned sessions are sorted to the top of the list for quick access.

## Why
Users with many sessions need a way to keep important ones easily accessible without scrolling or searching.

## How

**Backend:**
- New `pinned_sessions` table (migration 007) with `(user_id, session_id)` PK and org_id FK
- `pin_session`/`unpin_session`/`list_pinned_session_ids` in both PostgreSQL and in-memory storage
- `PUT /v1/sessions/{session_id}/pin` and `DELETE /v1/sessions/{session_id}/pin` endpoints
- `is_pinned` field on Session, populated when user context is available
- Requires authentication (returns 401 in anonymous mode)

**Frontend:**
- Pin/unpin API client functions and React Query hooks
- Pin toggle button (Pin/PinOff icons) on each SessionCard
- Pinned sessions sorted to top of list on sessions page

## Risk
- Low
- New table + endpoints only, no changes to existing data. `is_pinned` uses `skip_serializing_if` so omitted when not populated (backward compatible).

## Checklist
- [x] Tests added or updated (227 server unit tests pass, 700 core tests pass)
- [x] Backward compatibility considered (`is_pinned` omitted from JSON when None)

### DIFF
--- a/apps/ui/src/app/(main)/sessions/page.tsx
+++ b/apps/ui/src/app/(main)/sessions/page.tsx
@@ -1,7 +1,15 @@
 "use client";
 
 import { useState, useMemo } from "react";
-import { useAgents, useHarnesses, useSessions, useCreateSession, useLlmModels } from "@/hooks";
+import {
+  useAgents,
+  useHarnesses,
+  useSessions,
+  useCreateSession,
+  useLlmModels,
+  usePinSession,
+  useUnpinSession,
+} from "@/hooks";
 import { useRouter } from "next/navigation";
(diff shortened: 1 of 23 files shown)


chaliy:
## 📸 UI Screenshot

**Sessions page with pin icons on each session card (anonymous mode, pins not activated)**

Sessions page with pin icons on each session card (anonymous mode, pins not activated)

PR #527, opened 2026-02-19
fix(ui): render connection instructions as markdown

chaliy:
## What
Render `instructions_markdown` from connection providers as actual markdown instead of plain text. All markdown links now open in a new tab.

## Why
Connection provider instructions (e.g. Daytona) contain markdown syntax (links, bold, numbered lists) but were displayed as raw text in the API key dialog.

## How
- Replace plain `<div>` with `InlineStreamdownMessage` in the API key dialog
- Add `target="_blank" rel="noopener noreferrer"` to all Streamdown-rendered links via `components` prop override

## Risk
- Low
- All existing markdown rendering uses StreamdownMessage, so link behavior change is consistent globally

## Checklist
- [x] Tests added or updated

### DIFF
--- a/apps/ui/src/__tests__/connections-markdown.test.tsx
+++ b/apps/ui/src/__tests__/connections-markdown.test.tsx
@@ -1,0 +1,174 @@
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { ReactNode } from "react";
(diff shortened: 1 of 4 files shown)


chaliy:
## 📸 UI Screenshot

**Connect Daytona dialog with markdown-rendered instructions (links, bold, numbered list)**

Connect Daytona dialog with markdown-rendered instructions (links, bold, numbered list)

PR #796, opened 2026-03-12
feat(durable): add snapshot checkpointing for workflow event replay

chaliy:
## What
Add periodic snapshot checkpointing for workflow event replay. On replay, the engine loads the latest snapshot + events since that snapshot sequence number instead of replaying all events from the beginning.

## Why
Long-running workflows accumulate many events. Without snapshots, every replay must process all events from the start, making replay cost O(total_events). Snapshots reduce this to O(checkpoint_interval).

Fixes EVE-86

## How
- Add opt-in `snapshot_state`/`restore_state` methods to the `Workflow` trait
- Add `save_snapshot`/`load_latest_snapshot`/`load_events_after` to `WorkflowEventStore` trait
- Implement for both `InMemoryWorkflowEventStore` and `PostgresWorkflowEventStore`
- Add snapshot DDL to `006_v0.8.5.sql` migration (durable_workflow_snapshots table)
- Wire snapshot save/restore into executor replay loop in `process_workflow`
- Configurable interval via `WORKFLOW_SNAPSHOT_INTERVAL` env var (default 1000)
- Register snapshot restore factories in `WorkflowRegistry` for type-erased restoration

## Risk
- Low
- Snapshots are opt-in (workflows must implement `snapshot_state`/`restore_state`). Workflows that do not implement these methods behave exactly as before (full replay). Fallback to full replay if snapshot restore fails.

## Checklist
- [x] Tests added or updated (10 new tests)
- [x] Backward compatibility considered (fully backward compatible, opt-in)
- [x] Migration consolidated into 006_v0.8.5.sql (rebased on main)

### DIFF
--- a/crates/durable/src/engine/executor.rs
+++ b/crates/durable/src/engine/executor.rs
@@ -27,6 +27,10 @@ pub struct ExecutorConfig {
 
     /// Whether to validate actions before persisting
     pub validate_actions: bool,
+
+    /// Snapshot interval: save a snapshot every N events.
+    /// Set to 0 to disable snapshotting.
+    pub snapshot_interval: i32,
 }
 
 impl Default for ExecutorConfig {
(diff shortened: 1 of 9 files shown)


chaliy:
Feature already merged via #794. This PR is now redundant.

PR #822, opened 2026-03-13
chore(ci): add SDK compatibility checks

chaliy:
## What
- Add `.github/scripts/sdk-compat/resolve-sdk-versions.sh` plus per-language SDK smoke scripts for TypeScript, Python, and Rust, and wire `sdk-compat-matrix` / `sdk-compat-test` into `.github/workflows/ci.yml`.
- Make the SDK matrix resolver portable on stock GitHub runners by removing the `rg` dependency from the new CI path.
- Update the leased-resources spec link to the current migration filename on `main`.

## Why
- Catch API and SDK compatibility regressions against older published SDK releases before merge.
- Ensure the new SDK compatibility workflow actually runs on a fresh GitHub runner without assuming extra tooling.
- Keep the spec reference aligned with the migration layout now on `main`.

## How
- Resolve versions from `everruns/sdk` Git tags at workflow runtime and fan out a non-fail-fast matrix across languages and compatibility cohorts.
- Reuse the `build-binaries` server artifact, start the server in `DEV_MODE`, and run each language smoke script against the live API.
- Replace `rg` with `grep` in the resolver and point the leased-resources spec at `crates/server/migrations/010_v0.8.6.sql`.

## Risk
- Medium
- Risk is limited to CI behavior: tag resolution, dependency installation, or smoke-test assumptions can still make the workflow fail if runner assumptions are wrong.
- Verified locally with `./.github/scripts/sdk-compat/resolve-sdk-versions.sh`, `just pre-push`, `just pre-pr`, and a fresh-DB `sqlx migrate info/run` against PostgreSQL on port `51232`.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/.github/scripts/sdk-compat/resolve-sdk-versions.sh
+++ b/.github/scripts/sdk-compat/resolve-sdk-versions.sh
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# Keep this script portable to stock GitHub runners; avoid non-default tools.
 repo_url="https://github.com/everruns/sdk.git"
 
 mapfile -t versions < <(
   git ls-remote --tags --refs "$repo_url" \
     | awk '{print $2}' \
     | sed -E 's#refs/tags/v##' \
-    | rg '^[0-9]+\.[0-9]+\.[0-9]+$' \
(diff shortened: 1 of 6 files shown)


chaliy:
Superseded by #824 — rebased on main, fixed all 12 SDK compat failures, and added backward-compat server change.

PR #947, opened 2026-03-20
fix(ui): remove chat picker inset chrome

chaliy:
## What
Remove the inner native-looking gray field from the chat composer model and reasoning pickers.

## Why
The composer chip was rendering a second inset control inside the outer shell, which made the picker look visually broken.

## How
Use Base UI's non-native trigger path for the inline composer selects so the trigger stops painting native button chrome inside the chip. Added a regression test that asserts both inline composer selects use the non-native trigger configuration.

## Risk
- Low
- Limited to the chat composer model/reasoning selects. Main risk is keyboard/accessibility behavior regression, which was covered by using Base UI's built-in non-native trigger path and live smoke testing.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/apps/ui/src/__tests__/chat-panel.test.tsx
+++ b/apps/ui/src/__tests__/chat-panel.test.tsx
@@ -8,12 +8,11 @@ const mockSelectTrigger = jest.fn(
     children,
     nativeButton: _nativeButton,
     render: _render,
-    ...props
   }: {
     children?: React.ReactNode;
     nativeButton?: boolean;
     render?: React.ReactElement;
-  }) => <div {...props}>{children}</div>,
+  }) => <div>{children}</div>,
 );
 
 const mockSessionContext = {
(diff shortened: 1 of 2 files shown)


chaliy:
Addressed the mock-prop-spread note in commit a25cdd28. For the remaining context-reset thread: the shared test state is reset in the existing \ for this describe (, , and  are reset before each test in [apps/ui/src/__tests__/chat-panel.test.tsx](apps/ui/src/__tests__/chat-panel.test.tsx)). No additional code change needed there.

chaliy:
Addressed the mock-prop-spread note in commit `a25cdd28`.

For the remaining context-reset thread: the shared test state is already reset in the existing `beforeEach` for this describe. `chatEvents`, `llmModel`, and `reasoningEffort` are reset before each test in `apps/ui/src/__tests__/chat-panel.test.tsx`, so no additional code change is needed there.

PR #956, opened 2026-03-20
fix(ui): constrain ScrollArea max-height overflow

chaliy:
## What
Constrain shared UI scroll areas so sections that rely on `max-h-*` stay clipped and scroll internally, instead of letting long content spill into the next section.

## Why
A long tools list in the LLM history viewer could render past the end of the Tools section and overlap the following Input Messages section.

## How
- hide overflow at the shared `ScrollArea` root
- let the viewport inherit the parent max height
- add a regression test for the `max-h-*` usage pattern

## Risk
- Low
- Shared `ScrollArea` behavior changes for callers that use `max-h-*`; explicit-height callers should continue to work.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/apps/ui/src/__tests__/scroll-area.test.tsx
+++ b/apps/ui/src/__tests__/scroll-area.test.tsx
@@ -1,0 +1,19 @@
+import { render, screen } from "@testing-library/react";
+import { ScrollArea } from "@/components/ui/scroll-area";
+
+describe("ScrollArea", () => {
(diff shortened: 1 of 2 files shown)


chaliy:
Addressed the Copilot review comments in f13f28d: added an inset focus ring so the keyboard indicator stays visible with , and tightened the regression test with an explicit viewport presence assertion.

PR #1011, opened 2026-03-21
feat(mcp): add auth modes and OAuth connection flow

chaliy:
## What
Add explicit MCP server auth modes (`none`, `api_key`, `oauth`) and the OAuth connection flow needed for MCP tool execution.

## Why
Some MCP servers need no auth, some need an org API key, and some need per-user OAuth. The previous implementation also needed follow-up fixes so session OAuth validation, metadata discovery, and popup return handling behave correctly.

## How
- wire `auth_mode` / `oauth_provider_id` through core, server, worker, proto, storage, and UI
- add MCP OAuth authorize/callback handling, registration/discovery helpers, and bearer-token tool caching
- return `connection_required` from MCP tool execution when a user OAuth token is needed
- update the MCP settings UI and inline chat connection flow for OAuth-backed MCP servers
- tighten the OAuth flow by validating mode/session inputs, preserving custom ports in resource discovery, supporting preconfigured OAuth clients without dynamic registration, and fixing popup `return_to` handling

## Risk
- Medium
- Touches MCP registration, OAuth callback flows, worker auth resolution, and chat connection UX

## Checklist
- [x] Tests added or updated

### DIFF
--- a/apps/ui/src/__tests__/mcp-servers-page.test.tsx
+++ b/apps/ui/src/__tests__/mcp-servers-page.test.tsx
@@ -32,6 +32,7 @@ describe("McpServersPage", () => {
           url: "https://learn.microsoft.com/api/mcp",
           transport_type: "http",
           status: "active",
+          auth_mode: "api_key",
           api_key_set: false,
           headers: {},
           created_at: "2024-01-01T00:00:00Z",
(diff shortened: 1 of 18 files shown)


chaliy:
Addressed the latest review pass in `883ca74`:

- reject inconsistent MCP auth-mode/api-key combinations and preserve legacy `api_key_set` servers when `settings.auth_mode` is missing
- include MCP OAuth providers for non-admin users by reading org-scoped MCP rows directly
- harden OAuth discovery/registration/token exchange with safer issuer/origin derivation, SSRF validation, IPv6-safe origin formatting, and explicit non-2xx handling
- fix the chat popup query param to use `popup=true`
- require an API key in the MCP server create dialog when `auth_mode=api_key`
- keep worker gRPC auth-mode backward compatible when older servers omit `auth_mode`
- update the MCP executor auth-header comment for clarity

chaliy:
Superseded by new PR with compilation fixes and review feedback addressed.

PR #1039, opened 2026-03-22
feat(server): resolve connections from agent identity on session

chaliy:
## Summary
- When a session has `agent_identity_id`, connection resolution (tokens, installation IDs) now prefers `agent_identity_connections` over `user_connections`
- Falls back to user connections only when no identity connection exists for the requested provider
- `get_connection_user_for_session` returns `None` for identity-owned connections (no owning user)

## Changes
- **PostgreSQL queries** (`repositories/user_connections.rs`): Modified `get_connection_token_for_session`, `get_connection_user_for_session`, and `get_installation_id_for_session` with CTE-based priority resolution
- **In-memory implementation** (`memory/user_connections.rs`): Same identity-first logic for dev mode
- **Spec update** (`specs/agent-identities.md`): Documented identity-owned connection resolution

## Test plan
- [ ] Create agent identity with connection (e.g., GitHub API key)
- [ ] Create session with `agent_identity_id` set
- [ ] Verify tool execution uses identity connection instead of user connection
- [ ] Verify fallback: session without identity still uses user connections
- [ ] Verify fallback: session with identity but no matching provider falls back to user connections

### DIFF
--- a/crates/server/src/storage/connection_resolver.rs
+++ b/crates/server/src/storage/connection_resolver.rs
@@ -19,10 +19,12 @@ use super::backend::StorageBackend;
 use super::encryption::EncryptionService;
 use crate::auth::oauth::GitHubAppService;
 
-/// Resolves user connection tokens from stored user connections.
+/// Resolves connection tokens for tool execution.
 ///
-/// Tool execution still supports the existing session-based lookup
-/// (`session_id -> org_members -> user_connections`) for convenience.
+/// Session-based lookup priority:
+/// 1. If the session has an `agent_identity_id`, resolves from `agent_identity_connections`.
+/// 2. Falls back to `user_connections` via org membership.
+///
 /// Leased-resource cleanup additionally uses explicit owner-user lookups so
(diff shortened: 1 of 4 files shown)


chaliy:
All Copilot review feedback addressed in 0f02909:

1. **Credential shape mismatch (comments 1-4)**: Added `has_identity_conn` CTE (PostgreSQL) and `has_any` check (in-memory) that blocks fallback to user connections whenever the identity has **any** connection for the provider, regardless of whether it uses token or installation_id.

2. **Missing ORDER BY (comments 6-7)**: Added `ORDER BY uc.created_at ASC` to both `user_conn` CTEs for deterministic results consistent with `get_connection_user_for_session`.

3. **Spec/UI concern (comment 5)**: The identity connection management UI already exists: `identity-connections.tsx`, `identity-api-key-dialog.tsx`, `use-identity-connections.ts`, and the detail page at `agent-identities/[identityId]/page.tsx`. The spec edit is correct.

PR #1041, opened 2026-03-22
feat(deno): add Deno Deploy sandbox integration

chaliy:
## What
Add the Deno sandbox integration as a new execution capability and tighten the follow-up review fixes around cleanup metadata and connection handling.

## Why
Everruns needs a Deno-backed cloud sandbox option alongside Daytona. The previous patch also needed follow-up fixes before shipping: preserve non-secret Deno org routing metadata for cleanup, keep lease metadata explicitly non-secret, and resolve clippy/review gaps so the integration is merge-ready.

## How
- Add the new `integrations/deno` crate with client, connection provider, state helpers, tools, tests, and spec docs.
- Register the capability/connection provider in the workspace and wire Deno leased-resource cleanup into the worker.
- Persist the optional Deno org slug in sandbox state + lease metadata so cleanup can reconnect without depending only on ambient env state.
- Keep the UI-visible lease metadata limited to non-secret routing/debug fields and return `connection_required("deno")` instead of prompting for secrets in chat.
- Fix clippy feedback in the Deno client session-close paths.

## Risk
- Medium
- New remote execution provider touches tool execution, credential resolution, leased-resource cleanup, and worker cleanup paths.
- Live Deno API validation still depends on operator credentials (`DENO_DEPLOY_TOKEN` and, for personal tokens, `DENO_DEPLOY_ORG`).

## Security
- Threat categories reviewed: TM-TOOL, TM-API, TM-DOS, TM-AGENT-016, TM-API-015, TM-DENO
- Findings and resolutions:
  - Ensured Deno tokens are still resolved from user connections or operator env fallback, never persisted in sandbox state or lease metadata.
  - Stored only non-secret routing/debug metadata (`region`, optional `org`, workspace path, timestamps) in leased resources because that metadata is exposed via API/UI.
  - Kept `connection_required("deno")` behavior so the UI asks for connections instead of encouraging plaintext secret entry in chat/events.
  - Preserved explicit sandbox TTL behavior so Everruns never relies on Deno's websocket-scoped `session` lifetime.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1516,6 +1516,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "everruns-integrations-deno"
+version = "0.8.7"
+dependencies = [
+ "async-trait",
+ "base64",
+ "chrono",
+ "everruns-core",
+ "futures-util",
+ "http",
+ "inventory",
+ "reqwest 0.13.1",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tokio-tungstenite 0.26.2",
+ "tracing",
+ "urlencoding",
+ "wiremock",
+]
+
+[[package]]
 name = "everruns-integrations-docker"
 version = "0.8.7"
 dependencies = [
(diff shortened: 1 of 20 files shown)


chaliy:
Superseded by #1047 (rebased onto latest main, merge conflicts resolved, Copilot review comments addressed).

PR #1069, opened 2026-03-23
fix(server): include initial_files in agent upsert SQL query

chaliy:
## What
Fix PostgreSQL upsert_agent SQL query to include initial_files in INSERT, ON CONFLICT UPDATE SET, and RETURNING clauses.

## Why
The initial_files column was silently dropped during agent upsert (agents create --file or PUT /v1/agents/{id}), causing agent workspace files (AGENTS.md, specs, skills) to drift out of sync after an upsert. The in-memory backend handled this correctly, masking the bug in unit tests.

## How
Added initial_files to the three missing positions in the SQL query and the corresponding .bind() call. Added both a unit test (in-memory) and a PostgreSQL integration test proving the fix.

## Risk
- Low
- Only changes the upsert SQL query for agents. No schema change. The column already existed and was used by create/update paths.

## Security
- Threat categories reviewed: TM-SQL (database query change), TM-API (PUT endpoint behavior)
- The change adds a column that was already present in other queries (create, update). No new input surface. The initial_files value is bound via parameterized query (no injection risk). No auth/authz changes.

## Checklist
- [x] Tests added or updated (unit test + PostgreSQL integration test)
- [x] Backward compatibility considered (additive fix, no breaking change)
- [x] All review comments addressed (added PostgreSQL integration test per Copilot feedback)

Fixes EVE-179

### DIFF
--- a/crates/server/src/services/agent.rs
+++ b/crates/server/src/services/agent.rs
@@ -735,4 +735,40 @@ mod tests {
 
         assert_eq!(err.to_string(), "Capability not found");
     }
+
+    #[tokio::test]
+    async fn upsert_updates_initial_files() {
+        let db = Arc::new(StorageBackend::in_memory());
+        let service = AgentService::new(db.clone());
+        let caller = Caller::internal(DEFAULT_ORG_ID);
+
+        // Create agent with initial_files via upsert
+        let public_id = "agent_00000000000000000000000000000050";
+        let mut req = build_create_request(None);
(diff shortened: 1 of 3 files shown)


chaliy:
Addressed Copilot review: added a PostgreSQL repository-level integration test (test_agent_upsert_initial_files in repository_integration_test.rs) that exercises the actual SQL upsert_agent query with a create-then-update flow, verifying initial_files is persisted and updated correctly at the database level.

PR #1101, opened 2026-03-25
feat(server): add Prometheus /metrics endpoint

chaliy:
## Summary

- Expose application metrics in Prometheus exposition format at `GET /metrics` using `metrics` + `metrics-exporter-prometheus` crates
- Gauges bridged from existing `MetricsCollector` (workflows, tasks, workers, load, DLQ), cumulative counters by status, and histograms for HTTP request duration, LLM request duration, and tool execution duration
- **Metrics always enabled by default** — production isolation via `METRICS_ADDR` (separate internal-only server), not by disabling

## Production Safety (SaaS)

Metrics are **always on** (they're most valuable in production). To keep them off the public interface:

Set `METRICS_ADDR=127.0.0.1:9090` — this spawns a dedicated lightweight HTTP server bound to loopback that **only** serves `/metrics`. External traffic never reaches it. Scrapers access via pod-local networking, sidecar, or service mesh.

When `METRICS_ADDR` is unset, `/metrics` is mounted on the main API server (convenient for dev/local).

## Metrics

| Type | Metrics |
|------|---------|
| Gauges | `everruns_workflows_running`, `everruns_workflows_pending`, `everruns_tasks_pending`, `everruns_tasks_claimed`, `everruns_workers_active`, `everruns_load_ratio`, `everruns_dlq_size` |
| Counters | `everruns_tasks_total{status}`, `everruns_workflows_total{status}` |
| Histograms | `everruns_http_request_duration_seconds{method,path,status}`, `everruns_llm_request_duration_seconds{provider,model}`, `everruns_tool_execution_duration_seconds{tool}` |

## Env Vars

| Variable | Default | Description |
|----------|---------|-------------|
| `METRICS_ENABLED` | `true` | Enable/disable metrics collection |
| `METRICS_PREFIX` | `everruns` | Metric name prefix |
| `METRICS_ADDR` | *(unset)* | Dedicated bind address for internal-only metrics server (e.g. `127.0.0.1:9090`). Recommended for production. |

## Test plan

- [ ] `just pre-push` passes (verified locally)
- [ ] Start dev server, verify `curl localhost:9301/metrics` returns Prometheus text
- [ ] Verify gauges populate after ~10s (MetricsCollector bridge interval)
- [ ] Make API requests and verify `everruns_http_request_duration_seconds` histogram appears
- [ ] Run an agent session and verify `everruns_llm_request_duration_seconds` and `everruns_tool_execution_duration_seconds` populate
- [ ] Set `METRICS_ADDR=127.0.0.1:9090`, verify `/metrics` is served on port 9090 and NOT on the main API port
- [ ] Verify `METRICS_ENABLED=false` disables the endpoint entirely


chaliy:
## Addressing Copilot review feedback

All 10 comments addressed in 63c78a0. Summary:

| # | Comment | Resolution |
|---|---------|------------|
| 1 | `_total` metrics emitted as gauge instead of counter | **Fixed.** Now uses `counter!().increment(delta)` with delta tracking from last-seen DB totals. Proper monotonic counters, `rate()` compatible. |
| 2 | Unmatched path fallback causes cardinality explosion | **Fixed.** Falls back to `"unmatched"` instead of raw URI path. |
| 3 | Tests read env without guarding against leakage | **Acknowledged.** Added comment noting the test reads real env. Low risk — `METRICS_ENABLED` is not typically set in CI. The test serves as an intentional canary. |
| 4 | Env vars table `\|\|` formatting | **Already fixed** in the second commit (spec was fully rewritten). No `\|\|` in current spec. |
| 5 | `layer()` vs `route_layer()` for MatchedPath availability | **Fixed.** Changed to `route_layer(from_fn(...))` so MatchedPath is available after route matching. |
| 6 | `METRICS_PREFIX` read but never used | **Fixed.** Removed `prefix` field and `METRICS_PREFIX` env var entirely. Metric names are fixed constants. |
| 7 | `get_points()` clones entire ring buffer | **Fixed.** Added `MetricsCollector::get_latest()` that reads only the most recent point via `VecDeque::back()`. |
| 8 | Spec markdown `\|\|` table formatting | **Already fixed** in second commit. |
| 9 | `_total` metrics listed under Gauges in spec | **Fixed.** Separated into distinct "Counters" section in spec. |
| 10 | Histograms table `\|\|` formatting | **Already fixed** in second commit. |

## Security review

- **TM-API**: New `GET /metrics` — no auth (Prometheus standard). No user data exposed. Production isolation via `METRICS_ADDR` (separate loopback server).
- **TM-DOS**: HTTP path labels use `MatchedPath` (route template) with `"unmatched"` fallback — prevents cardinality explosion. Counter deltas bounded by DB values.
- No injection surface, no sensitive data in metrics output.
- Dependencies: `metrics` 0.24 + `metrics-exporter-prometheus` 0.16 — widely-adopted Rust ecosystem crates.

PR #1106, opened 2026-03-26
fix(server): make Prometheus metrics correct under horizontal scaling

chaliy:
## Summary

- Make Prometheus metrics correct under horizontal scaling (multiple API server replicas)
- **Gauges from DB**: Task/workflow totals emitted as gauges (renamed from `_total` to plain names like `everruns_tasks_completed`). All replicas read the same shared DB — Prometheus deduplicates identical values via `instance` label.
- **Counters from local events**: HTTP requests, LLM requests, and tool executions counted per-instance. Each replica only counts its own work — `sum()` across instances gives true total. No double-counting.
- Added new counters: `everruns_http_requests_total`, `everruns_llm_requests_total`, `everruns_tool_executions_total`

## Problem

The previous implementation used `counter!().increment(delta)` bridged from DB cumulative totals. With N replicas, each independently computes the same delta from the shared DB, causing Prometheus to see N× the actual count.

## Solution

| Category | Source | Multi-replica behavior |
|----------|--------|----------------------|
| Gauges | DB via MetricsCollector | All replicas emit identical values. Prometheus deduplicates via `instance` label. |
| Counters | Local events only | Each replica counts its own work. `sum()` gives true total. |
| Histograms | Local observations | Each replica records its own latencies. |

## Security

No security-relevant changes. Same metrics surface, no new endpoints, no auth changes. Only changes metric types and naming for correctness.

## Test plan

- [ ] `just pre-push` passes (verified locally)
- [ ] Start dev server, verify `curl localhost:9301/metrics` returns gauges for DB totals and counters for local events
- [ ] Verify `everruns_tasks_completed` appears as gauge (not counter)
- [ ] Verify `everruns_http_requests_total` appears as counter after making requests
- [ ] Verify `everruns_llm_requests_total` and `everruns_tool_executions_total` appear after running an agent session

### DIFF
--- a/crates/server/src/api/prometheus.rs
+++ b/crates/server/src/api/prometheus.rs
@@ -11,10 +11,11 @@
 //      policy or ingress rules.
 // Decision: No auth on /metrics (Prometheus standard).
 // Decision: Horizontal scaling model:
-//   - Gauges from DB (safe to duplicate across replicas — Prometheus deduplicates
-//     identical values via `instance` label)
+//   - Gauges from DB (each replica reports the same logical value; Prometheus
+//     keeps one series per `instance`, so queries for cluster-level values
+//     should aggregate, e.g. `max without(instance)`)
 //   - Counters from local events only (each replica counts its own work — no
-//     double-counting across replicas)
+//     double-counting across replicas; use `sum without(instance)` for totals)
 //   - Histograms from local observations (naturally partitioned per instance)
 
 use axum::http::header;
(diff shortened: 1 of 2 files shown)


chaliy:
## Addressing Copilot review feedback

All 7 comments addressed in a1a6f42. They all flagged the same issue:

**Prometheus does not auto-deduplicate gauges across instances.** Each `instance` label creates a separate time series.

Fixed all comments and spec docs to:
- Recommend `max without(instance)` for cluster-level DB-state gauge queries
- Recommend `delta()` instead of `increase()` for rate-like queries on gauges (since `increase()`/`rate()` are designed for counters and treat decreases as resets)
- Clarify that Prometheus keeps separate series per `instance` label

---

PR #1189, opened 2026-04-03
fix(core): delegate bash tool input_schema() to bashkit

chaliy:
## What
Replace hardcoded JSON schema in `BashTool::parameters_schema()` with `BASHKIT_TOOL.input_schema()`, extended with the everruns-specific `working_dir` parameter.

## Why
The hardcoded schema can drift from what bashkit actually supports when bashkit adds or changes parameters. Per the toolkit library contract (`specs/toolkit-library-contract.md`), consumers should delegate to `tool.input_schema()`.

Fixes EVE-239.

## How
- Added `TOOL_INPUT_SCHEMA` static that calls `BASHKIT_TOOL.input_schema()` and merges in `working_dir`
- Updated `execute_with_context` to read `commands` (bashkit field name) instead of `command`
- Updated all 40+ test callsites from `command` to `commands`
- Added `test_parameters_schema_delegates_to_bashkit` test verifying all bashkit properties and required fields are present

## Risk
- Low
- Agents sending `command` instead of `commands` will get an error. This is internal code with no backward compat requirement (per AGENTS.md).

## Security
- Threat categories reviewed: TM-BASH
- No new execution paths or trust boundary changes. Schema source moved from hardcoded to bashkit (single source of truth). Sandbox behavior unchanged.

## Checklist
- [x] Tests added or updated
- [x] Backward compatibility considered (internal code, no compat needed)

### DIFF
--- a/crates/core/src/capabilities/virtual_bash.rs
+++ b/crates/core/src/capabilities/virtual_bash.rs
@@ -74,17 +74,17 @@ static TOOL_SYSTEM_PROMPT: LazyLock<String> = LazyLock::new(|| {
 static TOOL_INPUT_SCHEMA: LazyLock<Value> = LazyLock::new(|| {
     let mut schema = BASHKIT_TOOL.input_schema();
     // Add everruns-specific working_dir param only if bashkit does not already define it.
-    if let Some(props) = schema.get_mut("properties").and_then(|p| p.as_object_mut()) {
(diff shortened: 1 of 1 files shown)


chaliy:
Both Copilot suggestions applied in bec4394:
1. Removed redundant default value from description text
2. Added contains_key guard so bashkit changes are not silently masked

PR #1191, opened 2026-04-04
feat(core): add bashkit native cancellation support

chaliy:
## What
Wire bashkit cancellation_token() to the timeout path in BashTool, collecting partial output instead of discarding it on timeout.

## Why
When bash execution times out, we drop the tokio future and lose all partial stdout/stderr. bashkit supports graceful cancellation via an AtomicBool token — setting it aborts at the next command boundary. This gives the LLM visibility into what ran before the timeout.

Fixes EVE-241.

## How
- Grab `bash.cancellation_token()` before exec
- Add a second unbounded channel to collect partial output chunks from the streaming callback
- On timeout, set the cancellation token and drain the partial channel
- Return sanitized partial output in the timeout error message (cleaned + priority-aware truncated)
- Added `collect_partial_output()` helper

## Risk
- Low
- Existing timeout behavior preserved when no partial output exists
- Partial output goes through the same sanitization pipeline as normal output

## Security
- Threat categories reviewed: TM-BASH, TM-DOS
- No sandbox boundary changes. Partial output sanitized via clean_exec_output + priority_aware_truncate before inclusion in error. Unbounded channel mirrors existing streaming channel.

## Checklist
- [x] Tests added or updated (74 existing bash tests pass)
- [x] Backward compatibility considered (timeout without partial output unchanged)

### DIFF
--- a/crates/core/src/capabilities/virtual_bash.rs
+++ b/crates/core/src/capabilities/virtual_bash.rs
@@ -235,15 +235,17 @@ impl Tool for BashTool {
         // Stream output via tool.output.delta events for live UI/CLI rendering.
         // bashkit's exec_streaming calls OutputCallback with (stdout_chunk, stderr_chunk)
         // after each command completes. We bridge to async emit via a channel.
-        // A second channel collects partial output for cancellation recovery.
+        // A bounded channel collects partial output for cancellation recovery
+        // without allowing unbounded memory growth.
         let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel::<(String, String)>();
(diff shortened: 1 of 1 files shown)


chaliy:
All 4 Copilot comments addressed in 275e74e:

1. **Unbounded channel** — Switched to bounded(128) + try_send, drops chunks when full.
2. **Misleading comment** — Reworded to describe in-flight cancellation, not instance reuse.
3. **Missing timeout test** — Acknowledged. Timeout tests for bashkit require long-running commands or injected delays, which are impractical in fast unit tests. The existing behavior (generic timeout error) is preserved when no partial output exists, so the regression risk is low.
4. **stdout/stderr separator** — Separated into two buffers, joined with the same "--- stderr ---" delimiter used elsewhere.

PR #1192, opened 2026-04-04
feat(core): wire session locale to bashkit builder

chaliy:
## What
Add `locale` field to `ToolContext` and pass it as `LANG` env var to the bashkit `Bash` builder.

## Why
bashkit supports locale-aware behavior but everruns never passes the session locale to it. The `BASHKIT_TOOL` static is built once with no locale, and the per-execution `Bash::builder()` doesn't set LANG.

Fixes EVE-242.

## How
- Added `locale: Option<String>` to `ToolContext` (with `None` in all constructors)
- In `BashTool::execute_with_context`, resolve locale from context (defaulting to en-US) and pass as `LANG` env var to `Bash::builder()`
- BashToolBuilder metadata (description, system_prompt) remains en-US since the Tool trait does not support per-session metadata — that requires a deeper refactor

## Risk
- Low
- Default behavior unchanged (LANG defaults to en-US)
- New field is `Option<String>` initialized to `None` everywhere

## Security
- Threat categories reviewed: TM-BASH
- LANG env var is set from session locale, not user input. No command injection risk. No sandbox boundary changes.

## Checklist
- [x] Tests added or updated (74 existing bash tests pass)
- [x] Backward compatibility considered (new field defaults to None)

### DIFF
--- a/crates/core/src/capabilities/virtual_bash.rs
+++ b/crates/core/src/capabilities/virtual_bash.rs
@@ -1290,6 +1290,38 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_bash_lang_env_default() {
+        let (context, _) = create_context_with_mock_store();
+        let tool = BashTool;
+
+        // Default locale (None) should set LANG to en-US
+        let result = tool
+            .execute_with_context(json!({"commands": "echo $LANG"}), &context)
+            .await;
+        if let ToolExecutionResult::Success(output) = result {
+            assert_eq!(output["stdout"], "en-US\n");
+        } else {
+            panic!("Expected success");
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


chaliy:
Both Copilot comments addressed in 628caff:

1. **BCP 47 vs POSIX locale** — bashkit is a pure-Rust virtual bash interpreter, not linked to libc. The LANG env var is simply exposed to scripts via $LANG, not consumed by locale-aware C functions. BCP 47 is the session-level format and passing it through is correct. No format conversion needed.
2. **Missing tests** — Added test_bash_lang_env_default (None -> en-US) and test_bash_lang_env_from_context_locale (uk-UA).

PR #1306, opened 2026-04-16
feat(runtime): expose shared turn context

chaliy:
## What
- add a shared `everruns_core::assemble_turn_context(...)` API and `AssembledTurnContext`
- refactor `ReasonAtom` and `everruns-runtime` to use the same context assembly path
- expose `InProcessRuntime::load_context(session_id)` for embedders
- add a public `inspect_context` example and runtime coverage
- document the embedder-facing context/custom-backend flow

## Why
The public in-process runtime could execute turns, but context assembly was still effectively hidden inside host-specific paths. That left embedders without a supported way to inspect or reuse the merged harness/agent/session context, and it let the worker-backed and in-process paths drift.

## How
- move turn-context assembly into `everruns-core`
- have `ReasonAtom` consume the shared assembled context instead of reloading and rebuilding its own view
- have `everruns-runtime` reuse that same API and surface it directly through `load_context`
- add docs and examples that show how to inspect the resolved model, tools, and message history

## Risk
- Medium
- Context assembly now sits on the hot path for both worker-backed and in-process reasoning. The main failure mode would be behavior drift in model resolution, message filtering, or blueprint/runtime-agent construction. Validation focused on those surfaces.

## Security
- Threat categories reviewed: TM-TOOL, TM-LLM, TM-FS, TM-DOS
- Findings and resolutions:
  - Reviewed shared tool/capability assembly to ensure the new core helper preserves capability dependency resolution and tool exposure semantics.
  - Reviewed model/context assembly to ensure the same controls/session/overlay/default precedence is preserved when resolving the active model.
  - Reviewed filesystem exposure to ensure `SystemPromptContext.file_store` remains an explicit injected dependency, not an implicit global.
  - Reviewed unbounded-context risk to ensure capability message filters and compaction config extraction still run through the merged effective overlay.
  - No new trust boundary was introduced; this change consolidates existing logic behind a public/shared API.

## Checklist
- [x] Tests added or updated

## Validation
- `just pre-push`
- `cargo run -p everruns-runtime --example inspect_context`
- `cargo test -p everruns-core -p everruns-runtime`
- `cargo test -p everruns-runtime`
- `cargo clippy -p everruns-core -p everruns-runtime -- -D warnings`
- `cargo check -p everruns-runtime --example in_process_runtime --example inspect_context`
- `cd apps/docs && npx -y node@22.12.0 scripts/generate-og-image.mjs && npx -y node@22.12.0 ./node_modules/astro/bin/astro.mjs build`

## Notes
- The docs toolchain now requires Node >= 22.12.0; local validation used `npx node@22.12.0` because the default shell runtime in this environment is Node 20.20.0.


chaliy:
Addressed the Copilot review follow-ups in `a54f9929`:

- `run_turn` now reuses the already-assembled first-turn context via `ReasonAtom::execute_with_assembled_context(...)`, avoiding duplicate message load + `RuntimeAgent` build on the first reason step.
- `load_context(session_id)` now uses the inspection path and works on empty sessions before the first user message.
- shared context assembly now collects message filters once and reuses them for query + post-load filtering.
- docs, example, spec, and runtime tests now cover pre-turn context inspection explicitly.

Validation after the follow-up change:
- `cargo test -p everruns-core -p everruns-runtime`
- `cargo clippy -p everruns-core -p everruns-runtime -- -D warnings`
- `cargo check -p everruns-runtime --example inspect_context`
- `just pre-push`

PR #1366, opened 2026-04-19
feat(ui): add chats view for owned sessions

chaliy:
## What
Add a top-level `Chats` navigation item between Dashboard and Sessions, backed by the existing sessions UI and a new owner-scoped sessions filter.

## Why
This is the first step in splitting conversational chats from broader session records used for debugging, tracing, and review.

## How
- add a new `/chats` page that reuses the shared sessions list client
- add `owned_by_me=true` to `GET /v1/sessions`
- persist `created_by` on newly created sessions and global chat sessions
- update nav, search, proxy protection, tests, and specs to reflect the Chats/Sessions split

## Risk
- Medium
- Can affect session listing behavior, auth-scoped filtering, and top-level navigation

## Security
- Threat categories reviewed: TM-API, TM-TENANT, TM-WEB
- Findings and resolutions:
  - TM-API: added an explicit boolean query param instead of inferring ownership from client-side state
  - TM-TENANT: owner filtering is still bounded by org scoping; `owned_by_me=true` returns empty when no authenticated user is present rather than broadening access
  - TM-WEB: `/chats` is protected by the same auth proxy path set as the rest of the main app; no new external data rendering paths were introduced
  - No new threat-model entries were needed; this change narrows reads rather than expanding cross-tenant or unauthenticated access

## Checklist
- [x] Tests added or updated

## Validation
- `just pre-push`
- `cargo test -p everruns-server test_list_sessions_filters_by_creator -- --nocapture`
- `cd apps/ui && npm test -- --runTestsByPath src/__tests__/sidebar.test.tsx src/__tests__/middleware.test.ts`
- Smoke test: started local dev app, registered a user, created a session, verified `/api/v1/sessions?owned_by_me=true` returned that session, and confirmed `/chats` resolved through the authenticated app route

### DIFF
--- a/apps/ui/src/__tests__/middleware.test.ts
+++ b/apps/ui/src/__tests__/middleware.test.ts
@@ -38,6 +38,7 @@ describe("auth proxy", () => {
       "/apps/:path*",
       "/capabilities/:path*",
       "/chat/:path*",
+      "/chats/:path*",
       "/dashboard/:path*",
       "/durable/:path*",
       "/evals/:path*",
--- a/apps/ui/src/__tests__/sidebar.test.tsx
+++ b/apps/ui/src/__tests__/sidebar.test.tsx
@@ -150,6 +150,8 @@ describe("Sidebar", () => {
 
     expect(screen.getByText("Chat")).toBeInTheDocument();
     expect(screen.getByText("Dashboard")).toBeInTheDocument();
+    expect(screen.getByText("Chats")).toBeInTheDocument();
+    expect(screen.getByText("Sessions")).toBeInTheDocument();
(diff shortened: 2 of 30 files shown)


chaliy:
Addressed the Copilot review point in  by adding a Postgres-backed repository integration test for , alongside the missing  migration that caused the PostgreSQL CI failure.

chaliy:
Addressed the Copilot review point in `2bed7ac1` by adding a Postgres-backed repository integration test for `list_sessions(..., created_by=Some(user_id), ...)`, alongside the missing `sessions.created_by` migration that caused the PostgreSQL CI failure.

PR #1379, opened 2026-04-19
fix(ui): remove Azure OpenAI from org setup

chaliy:
## What
Remove Azure OpenAI from the organisation setup onboarding picker while leaving Azure provider support in the main provider settings.

## Why
New org setup should only offer the onboarding paths we want to support there. Azure OpenAI still exists as a general provider option, but it should not appear in the setup flow.

## How
Narrow the setup page provider type to `openai | anthropic`, remove the Azure setup option, add a focused setup-page regression test, and update the org-creation manual test cases to match the current setup flow.

## Risk
- Low
- Could only affect the org setup provider picker and its submit payload
- Azure OpenAI remains available from Settings > Providers

## Security
- Threat categories reviewed: TM-WEB
- Findings and resolutions: reviewed the UI-only diff for auth bypass, data exposure, and input handling regressions. No new security-relevant surface or findings. This change only removes one setup-page option and narrows the local provider union used by that page.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/apps/ui/src/__tests__/org-setup-page.test.tsx
+++ b/apps/ui/src/__tests__/org-setup-page.test.tsx
@@ -1,0 +1,83 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import type { ReactNode } from "react";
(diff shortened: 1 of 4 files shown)


chaliy:
Reviewed the Copilot summary. No actionable review threads or suggested changes were raised, so no code changes were needed.

PR #1382, opened 2026-04-19
fix(server): restore v0.8.16 migration history

chaliy:
## What
Restore the original `016_eval_case_result_metadata.sql` and `017_eval_artifacts.sql` migrations, remove the incompatible `016_v0.8.16.sql` squash, and add a regression test that locks the filenames and SQL bodies long-lived deployments already recorded in `_sqlx_migrations`.

## Why
`origin/main` still reproduces EVE-360: a database that previously applied migration 17 fails startup with `migration 17 was previously applied but is missing in the resolved migrations` after the v0.8.16 squash.

Fixes EVE-360.

## How
Revert the v0.8.16 migration squash back to the original 016/017 files so sqlx sees the same migration history existing environments already applied. Add a test that prevents future renames or SQL rewrites of those two files.

## Risk
- Medium
- Touches migration history. Fresh databases still apply the same schema changes, and existing databases keep the exact migration versions/checksums they already recorded.

## Security
- Threat categories reviewed: TM-SQL
- Findings and resolutions: No new SQL or trust-boundary behavior was introduced. The change only restores previously shipped migration filenames and SQL bodies so existing `_sqlx_migrations` rows continue to validate during startup.

## Checklist
- [x] Tests added or updated

## Validation
- `cargo test -p everruns-server --test migration_history_test`
- Reproduced on `origin/main`: applying old 001..017 history, then `sqlx migrate run` from the squashed v0.8.16 set fails with `migration 17 was previously applied but is missing in the resolved migrations`
- Verified this branch: `sqlx migrate run` succeeds against the same database
- `PORT_PREFIX=271 DATABASE_URL=postgresql://everruns@localhost:27132/eve360_repro just start-all --no-watch --no-ui`
- `curl http://127.0.0.1:27101/health`
- `just pre-push`


chaliy:
Reviewed the Copilot summary. No actionable issues were raised; the change set matches the intended migration-history restore.

PR #1407, opened 2026-04-21
fix(ci): scope live jobs to real integration changes

chaliy:
## What
Split `Detect Changes` into two `dorny/paths-filter` steps so the paid live-integration filters (`daytona`, `deno`, `browserless`, `e2b`) run with `predicate-quantifier: every` while the broad positive-only filters keep the default behavior.

## Why
The previous single filter step mixed positive globs with negated `!**/*.md` rules under the default matching mode, which caused unrelated pushes to `main` to set the live-job outputs true and fire paid external CI jobs unnecessarily.

## How
Move the positive-only filters into `core_filter`, add a dedicated `live_filter` step for the exclusion-based integration filters, and wire the job outputs so only the paid live jobs read from the exclusion-safe step.

## Risk
- Low
- If the new filter wiring is wrong, live integration jobs could be skipped when integration code changes or still run when they should not.

## Security
- Threat categories reviewed: TM-DOS
- Findings and resolutions:
  - The broken filter routing caused unnecessary paid external jobs to run on unrelated pushes, which expands avoidable resource consumption. The workflow now scopes those jobs to real integration changes only.
  - No new secrets, permissions, third-party actions, or auth surfaces were introduced.

## Checklist
- [ ] Tests added or updated

## Validation
- Reviewed `git diff origin/main...HEAD` and recent `Detect Changes` job logs for runs `24697942500` and `24697932336` to confirm the root cause.
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml"); puts "yaml ok"'`
- `git diff --check`
- `just pre-push`
- Smoke test skipped: workflow-only change, no runtime app/API behavior changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,20 +26,20 @@ jobs:
     name: Detect Changes
     runs-on: ubuntu-latest
     outputs:
-      rust: ${{ steps.filter.outputs.rust }}
-      daytona: ${{ steps.filter.outputs.daytona }}
-      deno: ${{ steps.filter.outputs.deno }}
-      browserless: ${{ steps.filter.outputs.browserless }}
-      e2b: ${{ steps.filter.outputs.e2b }}
-      sdk_compat: ${{ steps.filter.outputs.sdk_compat }}
-      ui: ${{ steps.filter.outputs.ui }}
-      docs: ${{ steps.filter.outputs.docs }}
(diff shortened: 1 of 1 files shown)


chaliy:
Reviewed the Copilot overview comment. No code change was needed because it only summarizes the diff and matches the implemented workflow fix.

PR #1408, opened 2026-04-21
fix(deno): resolve sandbox A records explicitly

chaliy:
## What
Resolve Deno sandbox websocket hosts via explicit `A` record lookup before falling back to the system resolver, add regression coverage for the selection logic, and document the behavior in the Deno integration spec.

## Why
GitHub-hosted runners fail the Deno live test because they can still receive `AAAA` answers for the sandbox host even though they have no outbound IPv6 connectivity. Reordering the system resolver output was not enough; the client needs an explicit IPv4 path when an `A` record exists.

## How
Add `hickory-resolver` to query sandbox `A` records directly, use those IPv4 socket addresses when available, preserve the old system-resolver path only as a fallback when no IPv4 answer exists, and cover the selection behavior with unit tests.

## Risk
- Low
- Affects only the Deno websocket dial path; when no IPv4 record exists the client still falls back to the existing resolver behavior.

## Security
- Threat categories reviewed: TM-DENO, TM-DOS
- Findings and resolutions: No new trust boundary. The change narrows address-family selection for the same TLS-protected Deno sandbox host and does not broaden network reachability or secret exposure.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -704,6 +704,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "chacha20"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


chaliy:
Rebased onto the latest main at 09d097b5.

Manual live sweep run 24700501383 on this head reaches the Deno sandbox API and fails with VERIFICATION_REQUIRED_FOR_SANDBOXES instead of the original IPv6 Network is unreachable error.

That confirms the GitHub-runner IPv6 connectivity bug is addressed in code. The remaining Deno failure is an external org/account entitlement blocker, not the websocket dial path.

chaliy:
Follow-up tracking: this merge fixed the original IPv6 transport failure from EVE-380, but it should not be treated as fully shipped yet. The merge-commit CI on main still has a red Deno live gate because the sandbox call now fails on VERIFICATION_REQUIRED_FOR_SANDBOXES rather than Network is unreachable. I reopened EVE-380 and created a separate Linear follow-up to track the remaining live-test blocker cleanly.

PR #1411, opened 2026-04-21
fix(ci): revert deno sandbox A-record resolution

chaliy:
## What
Revert merge commit `687ef736` by removing the explicit Deno sandbox IPv4 `A`-record resolution path and dropping the temporary `hickory-resolver` dependency.

## Why
`#1408` fixed the original IPv6 transport failure from `EVE-380`, but I merged it before the shipping bar was actually met. The impacted Deno live path was still red, so this PR is a corrective rollback to get back to the pre-merge code state while `EVE-381` tracks the remaining live-test blocker cleanly.

## How
Use `git revert` against `687ef736`, keep the revert isolated to the Deno integration files and lockfile, and preserve all later `main` commits.

## Risk
- Medium
- This restores the prior Deno websocket dial behavior, so Deno sandbox connectivity on IPv6-less runners will regress to the pre-`#1408` state until `EVE-381` is resolved.

## Security
- Threat categories reviewed: TM-DOS, dependency risk
- Findings and resolutions: No new security-relevant surface added. This change removes the explicit DNS resolver dependency and restores the previous dial path; no auth, data exposure, or input-validation behavior changed.

## Checklist
- [ ] Tests added or updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -704,17 +704,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
-name = "chacha20"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
-dependencies = [
- "cfg-if",
- "cpufeatures 0.3.0",
- "rand_core 0.10.1",
-]
-
-[[package]]
 name = "chrono"
 version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


chaliy:
Manual smoke result on revert branch : integration live sweep run  still fails the impacted Deno path.\n\nExact Deno failure:\n\n\n\nSo this revert cleanly restores the pre-#1408 behavior, but it is **not mergeable under the shipping bar** because the impacted Deno smoke test is red again. Keeping the revert PR open as evidence / fallback, but not merging it.

chaliy:
Manual smoke result on revert branch `eve-381-revert-deno-a-record-resolution`: integration live sweep run `24701653868` still fails the impacted Deno path.

Exact Deno failure:

```text
create sandbox: "Failed to connect to Deno sandbox websocket address [2602:f70f::1]:443: Network is unreachable (os error 101)"
```

So this revert cleanly restores the pre-#1408 behavior, but it is not mergeable under the shipping bar because the impacted Deno smoke test is red again. Keeping the revert PR open as evidence / fallback, but not merging it.

chaliy:
Superseded by #1417. The revert branch was smoke-tested against the impacted Deno live path and still failed with the same websocket transport issue, so this is not a viable rollback and should stay closed.

PR #1429, opened 2026-04-22
fix(container-sandbox): gate coding-container behind flag

chaliy:
## What
Add an explicit backend-only `container_sandbox` feature flag, gate the `coding-container` built-in harness on that flag, and add regressions covering both the enabled and disabled paths.

## Why
`coding-container` was provisioned unconditionally, but its `container_sandbox` capability was gated by an internal flag name that did not exist in `InternalFeatureFlags`. That created a broken state in dev: the harness existed, but the capability registry did not expose `container_sandbox`, so agents never received the sandbox toolset.

## How
- add `InternalFeatureFlags.container_sandbox`, with `FEATURE_CONTAINER_SANDBOX` as the primary env var and `FEATURE_DOCKER_CAPABILITY` accepted as a legacy alias
- only include `coding-container` in built-in harnesses when that flag is enabled
- add seed regressions proving the harness is hidden when disabled and valid when enabled

## Risk
- Low
- The change narrows availability instead of broadening it: environments without the flag no longer advertise a broken harness. Environments that enable the flag must still provide the existing sandbox infra.

## Security
- Threat categories reviewed: TM-TOOL, TM-AGENT
- Findings and resolutions: reviewed the flag/harness gating path for accidental tool exposure. The patch prevents exposing a container harness unless the environment explicitly enables it.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/crates/container-sandbox/src/lib.rs
+++ b/crates/container-sandbox/src/lib.rs
@@ -32,7 +32,7 @@ use tools::{
 inventory::submit! {
     IntegrationPlugin {
         experimental_only: false,
-        feature_flag: None,
+        feature_flag: Some("container_sandbox"),
         factory: || Box::new(ContainerSandboxCapability),
     }
 }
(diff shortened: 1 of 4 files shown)


chaliy:
Closing this in favor of shipping the dev-only SaaS enablement directly from the wrapper.

PR #1436, opened 2026-04-22
fix(api): return round-trippable image ids

chaliy:
## What
Return public `img_*` IDs from `POST /v1/images` and `GET /v1/images` instead of raw UUIDs.

## Why
The image list/upload responses were not round-trippable. Clients had to transform raw UUIDs into `img_*` IDs before calling get/delete/thumbnail routes, even though those routes already required the typed public ID format.

## How
Use `ImageId` in the image response DTOs and row mappers, add regression tests for serialization and HTTP round-tripping, and update the committed OpenAPI schema examples for the image ID fields.

## Risk
- Low
- Could affect clients that incorrectly depended on the leaked raw UUID format instead of the documented public image ID contract

## Security
- Threat categories reviewed: TM-API
- Findings and resolutions: The change removes raw storage UUID leakage from public image API responses and aligns all image endpoints on the same public identifier format. No auth, permission, or new input-surface changes.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/crates/server/src/api/images.rs
+++ b/crates/server/src/api/images.rs
@@ -24,7 +24,6 @@ use serde::{Deserialize, Serialize};
 use std::io::Cursor;
 use std::sync::Arc;
 use utoipa::ToSchema;
-use uuid::Uuid;
 
 // ============================================
 // Constants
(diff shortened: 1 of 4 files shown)


chaliy:
Superseded by #1438, which landed the EVE-384 fix on main while this branch was in flight.

PR #1779, opened 2026-05-09
fix(volumes): block internal git source URLs

chaliy:
### Motivation
- Prevent SSRF/internal-network exposure: generic Git volume sources previously accepted `https`, `ssh`, `git`, and scp-like `git@` targets and persisted them as pending sync jobs, which allows an authenticated org user to cause server/worker outbound connections to internal hosts.
- Align Git source validation with existing SSRF-safe URL policies used elsewhere so persisted `source_config` cannot point at localhost, RFC1918, link-local, or cloud metadata endpoints.

### Description
- Restrict generic Git sources to absolute `https://` URLs only by updating `validate_git_url` to require the `https` scheme and to reject other schemes and scp-style `git@` forms. (file: `crates/server/src/domains/volumes/commands.rs`)
- Enforce no inline credentials by rejecting any username/password in the parsed URL. (file: `crates/server/src/domains/volumes/commands.rs`)
- Reuse the shared SSRF-aware validator `validate_safe_url` to block localhost, loopback, RFC1918, link-local, metadata, and other non-public hosts at write-time. (file: `crates/server/src/domains/volumes/commands.rs`)
- Add a regression test `git_volume_rejects_non_public_hosts_and_non_https_schemes` that asserts unsafe hosts and non-HTTPS schemes are rejected. (file: `crates/server/src/domains/volumes/commands.rs` tests)

### Testing
- Added unit test `git_volume_rejects_non_public_hosts_and_non_https_schemes` and kept existing secret-bearing-URL tests; these are under `crates/server/src/domains/volumes/commands.rs`.
- Attempted to run the targeted test with `cargo test -p everruns-server git_volume_rejects_non_public_hosts_and_non_https_schemes -- --exact`; compilation and test startup progressed in this environment but the full test completion was not observed due to a long build in the runtime window, so please run the full server test suite or `just pre-push` locally/CI to confirm green CI.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69fe7fa81c2c832b99948da8c78de654)

### DIFF
--- a/crates/server/src/domains/volumes/commands.rs
+++ b/crates/server/src/domains/volumes/commands.rs
@@ -6,6 +6,7 @@ use super::types::{
 use super::{VOLUME_MANAGE, VOLUME_VIEW};
 use crate::domains::common::*;
 use everruns_core::Policy;
+use everruns_core::url_validation::validate_safe_url;
 use everruns_core::typed_id::VolumeId;
 use everruns_durable::UpdateField;
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


chaliy:
Closing as duplicate of #1781 — same Codex task ID and identical diff. Tracking the volumes Git-URL hardening on #1781.

---

PR #1785, opened 2026-05-09
fix(mcp): hide HTTP metadata from discovery

chaliy:
## What
Remove HTTP method and path fields from MCP `discover` operation results, and add regression coverage for focused and `all: true` discovery responses.

## Why
MCP discovery should expose model-facing command metadata, not HTTP transport details like `GET /v1/sessions/{session_id}/databases/{name}/schema`.

## How
The MCP `tool_discover` JSON serializer now omits `method` and `path` while keeping internal `CommandMeta` unchanged for read-only classification, tracing, and other internal uses. Tests assert those fields are absent for `list_agents`, `discover all`, and the session database schema operation.

## Risk
- Low
- MCP clients that depended on `discover` returning HTTP transport metadata would need to use the public API docs/OpenAPI surface instead. `query` and `execute` command invocation are unchanged.

## Security
- Threat categories reviewed: TM-MCP / tool catalog data exposure.
- Findings and resolutions: the change reduces model-facing metadata exposure by removing HTTP method and route template fields from discovery output. No auth, authorization, input parsing, SQL, dependency, or resource-limit behavior changed. No threat-model update needed.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

## Validation
- `cargo test -p everruns-server --test mcp_endpoint_test test_mcp_discover -- --nocapture`
- `just pre-push`
- `PORT_PREFIX=274 CARGO_INCREMENTAL=0 just start-dev --no-watch` plus MCP smoke: `discover` for `session database schema` returned `has_method: false` and `has_path: false`

### DIFF
--- a/crates/server/src/api/mcp_endpoint/mod.rs
+++ b/crates/server/src/api/mcp_endpoint/mod.rs
@@ -1230,8 +1230,6 @@ async fn tool_discover(
             "name": entry.name,
             "category": entry.category,
             "description": entry.description,
-            "method": entry.method,
-            "path": entry.path,
             "read_only": entry.read_only,
             "output_shape": entry.output_shape,
         });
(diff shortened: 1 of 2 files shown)


chaliy:
Reviewed Copilot's summary. It raised no actionable comments or review threads; no code changes needed.

PR #1796, opened 2026-05-09
fix(mcp): align OAuth expires_in with actual JWT lifetime

chaliy:
## What
The MCP OAuth token endpoint (`POST /v1/mcp/oauth/token`) reported `expires_in: 3600` from a hardcoded `MCP_ACCESS_TOKEN_LIFETIME_SECS`, but the JWT access token it issued used `JwtService::generate_access_token()` whose lifetime is configured by `AUTH_JWT_ACCESS_TOKEN_LIFETIME` (default 900s). This PR derives `expires_in` from the same source as the JWT `exp` claim so the two never drift.

Resolves EVE-460.

## Why
MCP clients (e.g. Codex) cached and reused tokens by `expires_in`, so a 3600 vs 900 mismatch made tokens look valid client-side while the server rejected them. The result was apparent auth-loss even with Postgres-backed OAuth client and refresh-token storage.

## How
- Added `JwtService::access_token_lifetime_secs()` so callers can read the configured access-token TTL without poking the private `JwtConfig`.
- Removed `MCP_ACCESS_TOKEN_LIFETIME_SECS` from `auth/mcp_oauth.rs` and used the new accessor in both grant flows (`authorization_code` and `refresh_token`).
- Updated `specs/mcp.md` to describe the contract (`expires_in` reports the configured lifetime — default 900s) and updated the example response.
- Added a regression test `mcp_expires_in_matches_jwt_exp_minus_iat` that fixes a non-default lifetime, generates an access token, decodes it, and asserts `JwtConfig.access_token_lifetime` ≡ `JwtService::access_token_lifetime_secs()` ≡ `claims.exp - claims.iat`. This is the value the OAuth handlers now feed into `expires_in`.

## Risk
- Low. The change narrows clients' cached lifetimes to the value the server actually honors, so tokens stop being treated as valid past their JWT `exp`. No public-facing field is removed; only the value reported by `expires_in` changes.
- Operators who previously relied on the implicit 3600 may want to set `AUTH_JWT_ACCESS_TOKEN_LIFETIME=3600` (called out in the spec).

## Security
- Threat categories reviewed: TM-AUTH (access-token validity), TM-MCP. The change tightens behavior — `expires_in` no longer overpromises validity. No new attack surface.
- Findings and resolutions: None.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated
- [x] Backward compatibility considered (response shape unchanged; only the numeric value of `expires_in` changes when `AUTH_JWT_ACCESS_TOKEN_LIFETIME` ≠ 3600)

---
_Generated by [Claude Code](https://claude.ai/code/session_01D31EyHMKtKtVa4x5GW4mL9)_

### DIFF
--- a/crates/server/src/auth/jwt.rs
+++ b/crates/server/src/auth/jwt.rs
@@ -181,6 +181,13 @@ impl JwtService {
     pub fn refresh_token_lifetime_secs(&self) -> i64 {
         self.config.refresh_token_lifetime.as_secs() as i64
     }
+
+    /// Get access token lifetime in seconds. Matches the value used for the
+    /// JWT `exp` claim, so OAuth `expires_in` can be derived from the same
+    /// source instead of a separate hardcoded constant.
+    pub fn access_token_lifetime_secs(&self) -> i64 {
+        self.config.access_token_lifetime.as_secs() as i64
+    }
 }
 
 /// Hash a token for database storage (using SHA-256)
--- a/crates/server/src/auth/jwt.rs
+++ b/crates/server/src/auth/jwt.rs
@@ -141,7 +141,7 @@ impl JwtService {
             access_token,
             refresh_token,
             token_type: "Bearer".to_string(),
-            expires_in: self.config.access_token_lifetime.as_secs() as i64,
+            expires_in: self.access_token_lifetime_secs(),
         };
 
         Ok((token_pair, jti))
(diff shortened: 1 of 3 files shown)


chaliy:
Closing as superseded by #1788, which landed the same `expires_in` alignment (and additionally fixed the refresh-token rotation race) before this PR's CI completed. EVE-460 is already MITIGATED on main as of `11b1d1c`.

---

PR #1805, opened 2026-05-11
fix(volumes): block unsafe git sync source urls

chaliy:
### Motivation
- A background task was performing server-side `git clone` of user-provided volume sources without applying the repository's SSRF/egress protections, allowing authenticated volume managers to cause outbound connections to localhost/private/internal targets.

### Description
- Restrict generic Git volume URLs to absolute `https://` URLs by changing `validate_git_url` to require `https` and return an error for non-HTTPS and scp-like forms.
- Apply the shared SSRF-safe check by calling `validate_safe_url` from `everruns_core::url_validation` to block localhost, RFC1918, link-local, and cloud metadata hosts at write-time.
- Disallow inline credentials and simplify the validation error message to require an absolute HTTPS URL for syncable volumes in `crates/server/src/domains/volumes/commands.rs`.
- Add a regression test `git_volume_rejects_internal_or_non_https_urls` that asserts `CreateVolume` rejects `http`, `ssh`, `git` transports, scp-like `git@host:...`, and loopback/localhost `https` URLs.

### Testing
- Added unit test `git_volume_rejects_internal_or_non_https_urls` in `crates/server/src/domains/volumes/commands.rs` to cover the regression.
- Attempted to run `cargo test -p everruns-server git_volume_rejects_internal_or_non_https_urls -- --nocapture`, but the build in this environment kicked off dependency/toolchain compilation and did not complete during the run (compile and test progress observed but not finished).
- CI should run the full test suite and verify the new test and existing tests pass before merge.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69feaa53c780832bac93bba93c22ef88)

### DIFF
--- a/crates/server/src/domains/volumes/commands.rs
+++ b/crates/server/src/domains/volumes/commands.rs
@@ -5,6 +5,7 @@ use super::types::{
 };
 use super::{VOLUME_MANAGE, VOLUME_VIEW};
 use crate::domains::common::*;
+use everruns_core::url_validation::validate_safe_url;
 use everruns_core::Policy;
 use everruns_core::typed_id::VolumeId;
 use everruns_durable::UpdateField;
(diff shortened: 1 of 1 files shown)


chaliy:
Closing as superseded.

The same fix landed on `main` two days earlier via #1781 (`bf6f9a4 fix(volumes): block internal git source URLs`). Comparing the two:

- `validate_git_url` in `main` already restricts to `https`, rejects inline credentials, and applies `validate_safe_url` for SSRF protection — exactly what this PR proposes.
- Test coverage in `main` (`git_volume_rejects_non_public_hosts_and_non_https_schemes`) covers the same URL set as this PR's `git_volume_rejects_internal_or_non_https_urls`, plus extra ranges (10.x, 169.254.169.254, ssh://, git://, scp-like `git@host:`). Rebasing this branch produces a conflict resolving to the same code.

Re: copilot-pull-request-reviewer comments — they apply to the version on main too. Worth noting for follow-up:

1. **DNS rebinding via nip.io / dynamic DNS**: `validate_safe_url` is explicitly a static check (docs already note "Complement with runtime DNS-pinned checks where possible"). A defense-in-depth runtime resolve-then-check in the sync worker is worth a separate issue.
2. **No re-validation at sync time**: stored URLs from older code paths could bypass the new check. A defensive `validate_git_url` call on `source_config.url` right before the worker's `git clone` is a small follow-up.
3. **Swallowed error reason**: minor; mapping `UrlValidationError` into the `BadRequest` body would help operators debug rejections.

Closing this PR; none of these are introduced by this change vs. main.

---

PR #1843, opened 2026-05-17
fix(memory): wire memory store for grpc workers

chaliy:
## What
Wires persistent memory storage through the external gRPC worker path so `remember`, `recall`, and `forget` tools can reach the server-backed `DbMemoryStore` in production-style deployments.

## Why
Memory-enabled agents in production could call `remember`, but the tool failed with `Memory store not available` because gRPC workers inherited the default `WorkerAdapters::memory_store()` implementation, which returned `None`.

## How
- Added internal worker gRPC methods for memory store lookup, create, recall, forget, and active-count operations.
- Implemented those server-side RPCs using the existing org-scoped `DbMemoryStore` backend.
- Implemented `MemoryStoreBackend` for the gRPC org adapter and exposed it from `GrpcWorkerAdapters`.
- Made `WorkerAdapters::memory_store()` required so new worker adapter implementations cannot silently omit memory support.
- Added regression coverage for the gRPC worker backend and a server gRPC memory create/recall round trip.

## Risk
- Medium
- Touches internal worker protocol and gRPC worker/server memory plumbing. A proto mismatch or conversion bug could affect memory tools in external-worker deployments. Existing direct/in-process workers keep using the existing DB memory backend.

## Security
- Threat categories reviewed: TM-DURABLE, TM-TENANT, TM-TOOL, TM-API, TM-DOS
- Findings and resolutions:
  - gRPC auth boundary remains unchanged and covered by existing `TM-DURABLE-002` mitigation; this adds methods behind the same internal worker service.
  - Org scoping delegates to `DbMemoryStore`, which is bound to internal `org_id` and validates default-store public org IDs before create/read operations.
  - Store and memory IDs are parsed as typed IDs; invalid IDs return `InvalidArgument` instead of falling through to storage queries.
  - Memory recall remains bounded by the existing storage limit clamp, so the new RPC does not introduce an unbounded result path.
  - No new dependencies, secrets, public HTTP endpoints, or logging of memory content were added.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

## Validation
- `just pre-push`
- `cargo clippy -p everruns-worker -p everruns-server --all-targets -- -D warnings`
- `cargo test -p everruns-worker grpc_worker_adapters_exposes_memory_store_backend`
- `cargo test -p everruns-server --lib test_memory_rpc_create_and_recall_round_trip`
- `PORT_PREFIX=642 just start-all --no-watch --no-ui` plus `curl -fsS http://localhost:64201/health`
- `PORT_PREFIX=642 just stop-all`


chaliy:
Copilot review generated no actionable comments; no code changes required.

PR #1862, opened 2026-05-18
fix(runtime): block symlink escapes in RealDiskFileStore

chaliy:
### Motivation
- The real-disk `SessionFileStore` performed only lexical containment checks and then used filesystem APIs that follow symlinks, allowing a symlink placed inside the workspace to escape and read or write files outside the configured root. 
- The `file_system` capability promises operations confined to `/workspace`, so the backend must enforce containment after symlink resolution to prevent sandbox escapes.

### Description
- Add `ensure_no_symlink_escape(&self, absolute: &Path) -> Result<()>` which walks path components under `self.root` and rejects any component whose `symlink_metadata` indicates a symlink. 
- Invoke the symlink-escape guard from all real-disk entry points (`read_file`, `write_file`, `delete_file`, `list_directory`, `stat_file`, `create_directory`) and during seed path handling so operations fail before any symlink-following I/O occurs. 
- Add a regression test `symlink_escape_rejected` that creates a symlink inside the temporary workspace pointing outside and verifies a `read_file` call is rejected with a symlink-related error. 

### Testing
- Ran `cargo fmt -- crates/runtime/src/real_disk.rs` which completed successfully. 
- Added an automated unit test `symlink_escape_rejected` in `crates/runtime/src/real_disk.rs::tests` and attempted `cargo test -p everruns-runtime symlink_escape_rejected -- --nocapture`; the run in this environment was blocked by an existing Cargo artifact lock so the test execution was not fully observed here and should be validated by CI. 
- The change is intentionally minimal and focused on pre-I/O symlink checks to preserve existing functionality and semantics for non-symlinked paths.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0a9e95e7d8832ba1f5f381038c9621)

### DIFF
--- a/crates/runtime/src/real_disk.rs
+++ b/crates/runtime/src/real_disk.rs
@@ -135,6 +135,47 @@ impl RealDiskFileStore {
         Ok(absolute)
     }
 
+    fn ensure_no_symlink_escape(&self, absolute: &Path) -> Result<()> {
+        if !absolute.starts_with(&self.root) {
+            return Err(AgentLoopError::tool(format!(
+                "path escapes workspace root: {}",
+                absolute.display()
+            )));
+        }
+
+        let rel = absolute.strip_prefix(&self.root).map_err(|_| {
+            AgentLoopError::tool(format!(
+                "path is outside workspace root: {}",
+                absolute.display()
+            ))
+        })?;
+
+        let mut current = self.root.clone();
+        for component in rel.components() {
+            if let Component::Normal(seg) = component {
+                current.push(seg);
+                match std::fs::symlink_metadata(&current) {
+                    Ok(metadata) => {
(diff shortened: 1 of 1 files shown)


chaliy:
Closing as superseded by #1876.

#1876 is the more thorough version of the same symlink-escape fix:
- Targets the current `SessionFileSystem` trait (this PR targets the older `SessionFileStore` trait which no longer exists on main after the runtime refactor — would need to be rewritten anyway).
- Adds the same `reject_symlink_path` guard on read/write/delete/list/stat/create/seed.
- Additionally skips symlink children in `list_directory` (defense in depth).
- Has 4 regression tests (read/list/write/skip-children) instead of 1.

Same TOCTOU caveat raised by Copilot applies to both PRs and will be addressed there.

---

PR #1941, opened 2026-05-22
feat(core): add fingerprint loop detection

chaliy:
## What
- Emit stable normalized fingerprints for tool calls and tool results on tool execution events.
- Use those fingerprints in the optional `loop_detection` capability to warn when the same tool call produces the same result repeatedly.
- Add loop detection to the Generic harness, preserve fingerprint metadata during runtime/server replay, and let ercode inherit the default iteration behavior instead of forcing 20 iterations.

## Why
Long agentic sessions need cycle detection that can catch repeated tool behavior without relying on a low hard max-iteration cap.

## How
- Added a core fingerprint helper that hashes normalized tool-call/result payloads and ignores volatile fields.
- Extended tool started/completed event data with optional fingerprint fields and taught `ActAtom` to populate them.
- Rehydrated fingerprints from recent tool-result message metadata for bounded rolling-window loop detection, with a normalized tool-call-batch fallback.
- Updated specs and focused tests for core, runtime, server harness registration, and ercode.

## Risk
- Medium
- What can break: repeated intentional identical tool calls may now trigger a loop warning, and consumers that strictly deserialize tool events must tolerate the new optional fields.

## Security
- Threat categories reviewed: TM-TOOL, TM-LLM, TM-DOS, TM-BASH.
- Findings and resolutions: no new tool execution path, auth/tenant boundary, secret handling, or unbounded history scan was added. The detector uses recent message metadata and compact hashes only; prompt changes are bounded to a warning message. No threat-model update was needed.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

## Validation
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p everruns-core loop_detection --lib`
- `cargo test -p everruns-core tool_fingerprint --lib`
- `cargo test -p everruns-core test_act_atom_uses_tool_call_hooks_for_execution_arguments --lib`
- `cargo test -p everruns-runtime tool_completed_replay_preserves_fingerprints_as_metadata --lib`
- `cargo test -p everruns-server test_generic_harness_has_expected_capabilities --lib`
- `cargo test -p everruns-server test_generic_harness_capabilities_are_registered --lib`
- `cargo test -p everruns-coding-cli coding_harness_enables_loop_detection`
- `cargo run -p everruns-coding-cli -- --provider sim --session-dir /tmp/ercode-loop-detection-smoke -C /tmp --print "say hello"`
- `just pre-push`


chaliy:
Final review note: I reviewed the Copilot low-confidence note about `connection_required` in `tool_result_fingerprint`. No code change is needed: the current expression compiles and passes clippy/CI, and the JSON macro serializes from the borrowed `ToolResult` without moving it out of the shared reference. Latest full PR CI is green.

PR #2077, opened 2026-06-06
fix(core): disclose generic tool search stubs

chaliy:
## What
Update generic `tool_search` deferral so each hidden parameter schema is replaced with a progressive-disclosure stub that names the exact `tool_search` query needed to reveal the full schema.

## Why
`DeferSchemaHook` hid parameter schemas but only left a generic hint. Deferred tools now point the model at the concrete search path for their full schema while preserving the real schema internally for `tool_search` results.

## How
- Make deferred stubs tool-specific and keep them as open JSON objects so provider-side calls remain compatible after schema lookup.
- Preserve full parameters in `full_parameters` as before.
- Extend focused tests to prove stubs expose the search hint and still retain the full schema for progressive disclosure.
- Sync `specs/tool-search.md` with the generic fallback behavior.

## Risk
- Low
- Affected surface is generic tool schema metadata sent to LLM providers. The stub remains a standard open JSON object and does not add provider-specific JSON Schema extension keys.

## Security
- Threat categories reviewed: TM-TOOL, TM-LLM.
- Findings and resolutions: no new execution authority, auth, tenant boundary, or secret handling. The change only adjusts model-facing schema descriptions for already-visible deferred tools. Full schemas remain process-local in `full_parameters` and are still returned through the existing `tool_search` path.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

Validation:
- `cargo fmt --check`
- `cargo test -p everruns-core capabilities::tool_search`
- `just pre-push`
- `bash scripts/lib/check-migration-ordering.sh`

### DIFF
--- a/crates/core/src/capabilities/tool_search.rs
+++ b/crates/core/src/capabilities/tool_search.rs
@@ -9,10 +9,11 @@
 // How it works:
 //   1. A `tool_definition_hook` (`DeferSchemaHook`) runs at runtime-agent build
 //      time. When the agent carries at least `threshold` tools, it replaces the
-//      parameter schema of every deferrable tool with a minimal stub. Only the
-//      name + description survive, so the model still sees that the tool exists
-//      but pays no token cost for its parameters. Tools marked
-//      `DeferrablePolicy::Never` (e.g. high-frequency tools) keep full schemas.
+//      parameter schema of every deferrable tool with a minimal disclosure stub.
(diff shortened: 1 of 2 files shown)


chaliy:
Reviewed Copilot's COMMENTED review. It only contains a summary and no actionable inline comments or requested changes, so there is nothing to change.

PR #2161, opened 2026-06-12
feat(core): open LLM driver model for embedder-defined providers

chaliy:
## What
Opens the LLM provider model so embedders can register and use their own LLM
providers without modifying core enums.

- `LlmProviderType` and `ProviderType` gain an `External(Arc<str>)` variant.
  Unknown provider ids parse to `External` instead of erroring (`FromStr` is
  now `Infallible`); serialization stays a plain wire string. External ids are
  normalized to lowercase so casing variance never yields duplicate providers.
- Driver factories now receive a single `DriverConfig` (provider type,
  optional api_key/base_url, and a new `ProviderMetadata`) instead of
  `(api_key, base_url)`, so providers can authenticate via OAuth tokens /
  account ids without another signature change.
- `DriverRegistry::register` panics on duplicate registration (silent
  overwrites hid bugs); `register_or_replace` and `register_external` are
  added for intentional overwrite and external registration.
- `create_driver` no longer requires an api_key for `External` providers
  (they may authenticate via metadata).
- `ModelWithProvider` and `ProviderConfig` carry the new metadata; worker/
  server provider-type parsing maps unknown non-empty ids to `External` so
  embedder-defined providers round-trip through gRPC and the database, and
  empty/whitespace ids fail fast as a config/store error.

## Why
EVE-561. Adding a provider previously meant editing closed enums in core and
threading a new factory signature everywhere. Embedders need to plug in custom
providers (e.g. OAuth-based gateways) without forking core.

## How
- Manual `Serialize`/`Deserialize` + `ToSchema`/`PartialSchema` for
  `LlmProviderType` (an open string at the wire level; `Arc<str>` has no
  derive support).
- `From<LlmProviderType> for ProviderType` and `From<&ModelWithProvider> for
  ProviderConfig` carry `External` and metadata through.
- All built-in driver factories (OpenAI/OpenRouter/Azure/Completions,
  Anthropic, Gemini, Bedrock, LlmSim) adopt the `DriverConfig` argument.
- Regenerated `docs/api/openapi.json`: `LlmProviderType` is now an open
  string schema.

## Risk
- Low / Medium. Behavior change: duplicate driver registration now panics at
  startup (intended). Unknown DB provider ids now resolve to `External`
  instead of erroring / silently falling back to llmsim.
- No backward-compatibility shims (internal code; no spec requires them).

## Security
- Threat categories reviewed: TM-LLM. Provider credentials still never cross
  the command-host trait boundary; `ProviderMetadata` (OAuth tokens, account
  ids) lives only inside driver construction, mirroring the existing api_key
  handling. No new credential exposure surface.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated (external registration, duplicate-panic,
  register_or_replace, external parsing, case-insensitivity, empty-rejection;
  all call sites updated)
- [x] Backward compatibility considered (internal-only; no shims needed)
- [x] Security review performed (TM-LLM; credentials stay in driver layer)
- [x] All review comments addressed

## Validation
- `cargo build --all-features --workspace` clean.
- Tests: core 2325, server lib 1483, worker lib 95, runtime lib 48, runtime
  integration 46 — all pass.
- `cargo fmt --check` clean; clippy clean on touched crates.
- OpenAPI spec freshness: regenerated; only the `LlmProviderType` schema
  changed.
- Three Copilot review comments addressed (external id casing normalization;
  empty `provider_type` rejection in DB and gRPC paths).

### DIFF
--- a/crates/anthropic/src/driver.rs
+++ b/crates/anthropic/src/driver.rs
@@ -963,8 +963,9 @@ impl std::fmt::Debug for AnthropicLlmDriver {
 /// register_driver(&mut registry);
 /// ```
 pub fn register_driver(registry: &mut DriverRegistry) {
-    registry.register(ProviderType::Anthropic, |api_key, base_url| {
-        let driver = match base_url {
+    registry.register(ProviderType::Anthropic, |config| {
+        let api_key = config.api_key.as_deref().unwrap_or("");
+        let driver = match config.base_url.as_deref() {
             Some(url) => AnthropicLlmDriver::with_base_url(api_key, url),
             None => AnthropicLlmDriver::new(api_key),
         };
--- a/crates/bedrock/src/driver.rs
+++ b/crates/bedrock/src/driver.rs
@@ -70,13 +70,13 @@ impl BedrockLlmDriver {
 
 /// Register the Bedrock driver with the given registry.
 pub fn register_driver(registry: &mut DriverRegistry) {
-    registry.register(
-        ProviderType::Bedrock,
-        |api_key, _base_url| match BedrockLlmDriver::new(api_key) {
+    registry.register(ProviderType::Bedrock, |config| {
+        let api_key = config.api_key.as_deref().unwrap_or("");
+        match BedrockLlmDriver::new(api_key) {
             Ok(driver) => Box::new(driver) as BoxedLlmDriver,
             Err(e) => Box::new(FailDriver(e.to_string())) as BoxedLlmDriver,
-        },
-    );
+        }
+    });
 }
 
 /// Driver that immediately fails with a credential error.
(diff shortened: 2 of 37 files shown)


chaliy:
Addressed all three review comments in `aa19ffd`:

1. **External id casing** — external provider ids are now normalized to lowercase at every construction point (`ProviderType::external`, `register_external`, and both `FromStr` impls), so the same id in different casing resolves to one provider. Registration and parsed lookup now agree case-insensitively, matching the built-in matching leniency. Added `test_external_provider_id_is_case_insensitive`.
2. **Empty `provider_type` in DB** — `parse_provider_type` now returns a configuration error for empty/whitespace instead of silently producing `External("")`. Added `parse_provider_type_rejects_empty_value`.
3. **Empty proto `provider_type`** — the gRPC `proto_model_with_provider_to_model` path now fails fast with a clear store error on empty input before parsing.

Non-empty unknown ids still map to `External` as intended. Core/server/worker lib tests green (2325 / 1483 / 95); fmt + clippy clean.

---

PR #2189, opened 2026-06-13
feat(workspace): fs move/copy/download, workspace_id, file rename

chaliy:
## What

Implements the three follow-ups left by #2140 (Workspace-as-durable-entity):

1. **Workspace fs parity** — add `move`, `copy`, and raw `download` to the
   canonical `/v1/workspaces/{workspace_id}/fs` surface:
   - `POST /fs/_/move`, `POST /fs/_/copy`
   - `GET /fs/_/download/{path}` (Content-Disposition: attachment)
   - `GET /fs/{path}` now also returns raw bytes for
     `Accept: application/octet-stream`
2. **`workspace_id` on the session response** — added to the core `Session`
   model (and the UI `Session` type, optional), backed by the
   `sessions.workspace_id` column, so clients read the attached workspace
   directly instead of deriving it from the session id via prefix swap.
3. **Rename** `SessionFileService` → `WorkspaceFileService` and the nine
   `*_session_file*` commands → `*_workspace_file*` (struct names + command
   identifiers, e.g. `list_session_files` → `list_workspace_files`).

OpenAPI regenerated; `specs/workspace.md` and a few spec references updated.

## Why

#2140 split Workspace out as a first-class durable entity but left the file
surface and naming half-migrated: the workspace fs lacked move/copy/download
(they lived only on the legacy session alias), the session response still
forced clients to derive the workspace id, and the internal types/commands
still read "session". These three close that gap.

## How

- The new workspace fs handlers resolve the workspace and call
  `WorkspaceFileService` directly (consistent with the existing workspace fs
  handlers), reusing the shared raw-file response helpers from the session
  routes (now `pub(crate)`).
- `Session.workspace_id` is populated through `row_to_session` from the
  `sessions.workspace_id` column, selected across all session projections.
  For the default 1:1 case it still equals the session id; shared workspaces
  (future) will diverge. The proto Session conversion reconstructs it from the
  session id under the same equality invariant.
- The rename is scoped to the service type and command structs/identifiers.
  The storage layer (`create_session_file`, `SessionFileRow`, the
  `domains/session_files` module path) and the core `SessionFile` DTO /
  `SessionFileSystem` trait are intentionally left as-is. The renamed commands
  still serve the legacy `/v1/sessions/{session_id}/fs/*` alias routes; HTTP
  paths are unchanged.

## Risk

- **Low.** No migrations, no schema change beyond reading the existing
  `sessions.workspace_id` column. New fs endpoints are additive; `workspace_id`
  is an additive response field. The one place that deserializes `Session` from
  JSON (proto conversion) was updated to populate it. The rename changes MCP
  tool identifiers for the file commands but no HTTP paths.

## Security

- **Threat categories reviewed: TM-API, TM-FS, TM-AUTHZ, TM-TENANT.**
- New fs endpoints enforce `WORKSPACE_MANAGE` (move/copy — writes, rejected on
  archived workspaces) and `WORKSPACE_VIEW` (download — read) before any DB
  call, identical to the existing workspace fs handlers. Workspace resolution
  goes through `db.get_workspace(org_id, …)`, so cross-org access is impossible
  at the storage layer (TM-TENANT).
- Path handling reuses the existing file-service validation (`..` traversal,
  null bytes, reserved `_/` prefix) and the existing Content-Disposition
  sanitization for downloads (ASCII fallback + `filename*`). No new traversal
  or injection surface.
- `workspace_id` exposure is read-only and adds no new data: the caller already
  owns the session, and the value is derivable from the session id today via
  the equality invariant.
- Command rename is mechanical; per-command policies (`SESSION_MANAGE` /
  `SESSION_VIEW`) are preserved. No new external surface, no new auth code, no
  secrets.

## Follow-ups

- **UI consumption of `workspace_id`** (stop deriving from session id) stays
  deferred until shared workspaces actually ship — the field is now exposed and
  typed (optional in TS) but the prefix-swap derivation still works under the
  1:1 invariant. Rationale: behavior-neutral until divergence is possible.
- **`domains/session_files` module path** kept as-is (only the type/commands
  renamed) to bound churn; can be renamed when the legacy session fs surface is
  fully retired.
- Workspace-scoped git / tables / KV / secrets remain open questions in
  `specs/workspace.md` (out of scope).

## Checklist

- [x] Tests added or updated (new workspace fs integration test; the workspace
  CRUD/fs routes are now wired into the test harness; command-policy +
  mcp-catalog tests updated for the new names)
- [x] Backward compatibility considered (legacy session fs routes unchanged;
  additive response field; no migrations)


chaliy:
Thanks @copilot-pull-request-reviewer — all review points addressed in 05f63cd:

- **Derive `workspace_id` from the session id in test/example/fixture sessions** (turn_based_execution, scoped_mcp, reason_atom_test, capabilities/session, platform_store): now `WorkspaceId::from_uuid(session_id.uuid())` so they honor the `workspace.id == session.id` equality invariant and don't mask mapping bugs. (The proto round-trip test was already derived in 44d44e2, which also fixed the matching production path in `proto_session_to_schema`.)
- **"session workspace root" → "workspace filesystem root"** on the move/copy/stat request path docs, since those schemas now back both the workspace and the legacy session surfaces. OpenAPI regenerated.
- **Wildcard path params** (workspace fs GET and `_/download`): descriptions now state that nested paths are passed as literal `/`-separated segments (not a single URL-encoded value), matching the axum `{*path}` route behavior.
- **specs/workspace.md**: corrected to say the whole `/fs/*` surface (path routes + `/fs/_/*` actions) is mirrored on the legacy session aliases, not just `/fs/_/*`.

Resolving the corresponding threads.

---

PR #2246, opened 2026-06-15
fix(core): block web_fetch redirects before DNS

chaliy:
### Motivation
- Close a DNS-layer policy bypass where fetchkit resolved redirect targets before the session `NetworkAccessList` was enforced, allowing DNS queries to disallowed hosts via redirect chains.
- Preserve fetchkit's resolve-then-check protections while ensuring per-hop ACL checks run before any DNS resolution for egress-routed fetches.

### Description
- Vendor a minimal local patch of `fetchkit` (0.4.0) adding an optional `url_policy` hook on `ToolBuilder` / `FetchOptions` so callers can evaluate URL policies before DNS pinning and per-hop transport calls. (adds `vendor/fetchkit` and a `[patch.crates-io]` entry). 
- Have `FetchOptions::validate_url` evaluate the `url_policy` and call it from redirect-following and single-hop code paths before DNS pinning (`fetchers/default.rs` and `tool.rs`).
- Wire the runtime session ACL into fetchkit for egress-backed `web_fetch` calls by injecting a `url_policy` built from `ToolContext.network_access` so every hop is policy-checked pre-DNS (`crates/core/src/capabilities/web_fetch/mod.rs`).
- Update `egress_transport` documentation to reflect the new pre-DNS injection point (`crates/core/src/capabilities/web_fetch/egress_transport.rs`).
- Add a regression test that simulates a redirect to a disallowed host and asserts the redirect target is rejected before a second egress transport call occurs (`crates/core` web_fetch tests).

### Testing
- Ran formatting and linting: `cargo fmt --check` and `cargo clippy -p everruns-core --all-targets --all-features -- -D warnings`, both passed.
- Ran the full focused web_fetch test suite: `cargo test -p everruns-core --all-features capabilities::web_fetch --no-fail-fast`, resulting in all `web_fetch` tests passing (60 passed, 0 failed).
- Executed the targeted regression test: `cargo test -p everruns-core --all-features test_egress_path_blocks_disallowed_redirect_before_transport --no-fail-fast`, which passed.
- All added/modified tests and the focused suite succeeded locally; no test failures were observed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a305105cde4832ba9eac2339bbc36ff)


chaliy:
Closing in favor of a lighter fix for the same threat class.

**Context.** This PR targets a real residual gap: in the post-#2139 architecture, fetchkit resolves DNS for a redirect target *before* the egress boundary denies the hop, so a disallowed host can be DNS-resolved via a redirect chain (a DNS side-channel — note the actual HTTP connection is still blocked at the egress boundary on every hop). The fix here closes it by **vendoring a patched fetchkit 0.4** (`vendor/fetchkit`, +21k lines, plus a `[patch.crates-io]` entry) that adds a pre-DNS `url_policy` hook.

**Why close.** The primary exfil vector — a redirect to an attacker-controlled cross-host target — is now closed by **#2249** (`same_host_redirects_only_if_set`, enabled when the deployment `SystemAllowlist` is active) using fetchkit's *existing* API, with no vendoring. Combined with the egress boundary already re-denying every hop (#2139) and #2239 keeping DNS pinning inside the egress boundary for the direct/lua path, the remaining exposure is a DNS-lookup side-channel limited to deployments with **no** system allowlist configured.

Vendoring an upstream crate (and owning a `[patch.crates-io]` fork that must be re-vendored on every fetchkit upgrade) is a significant, long-lived maintenance commitment that isn't justified by closing a low-severity DNS side-channel. If we want the stronger *pre-DNS per-hop ACL* guarantee even without a system allowlist, the right path is to land the `url_policy` hook **upstream in fetchkit** and consume it normally — happy to reopen/track that as a follow-up rather than carry a vendored fork.

Superseded by #2249 (+ #2239). Closing.

---

PR #2310, opened 2026-06-18
fix(agents): throttle LLM analysis

chaliy:
### Motivation
- The on-demand `analyze_agent` flow made paid utility-LLM calls built from the full authored/resolved prompt and tool list without input-size caps, per-user/org admission limits, or a concurrency guard, enabling cost/availability abuse. 
- Add lightweight, in-process controls so large or repeated analyses cannot drive provider spend, exhaust quotas, or tie up outbound connections while preserving the advisory functionality.

### Description
- Introduce analysis admission controls: per-org and per-caller rolling-window limits and a small global semaphore (`acquire_analysis_permit`) that `AnalyzeAgent` acquires before running analysis. 
- Add a pre-call input-size guard (`MAX_CHECKER_INPUT_BYTES`) that rejects oversized checker inputs before any external utility-LLM request is issued. 
- Add a `RateLimited` `CommandErrorKind` and surface it as HTTP 429 / gRPC resource-exhausted with a `retry_after_seconds` hint and machine `code` for callers. 
- Add a unit test asserting oversized analysis input is rejected and update the threat model with the mitigation (TM-DOS-018).

### Testing
- Ran `cargo fmt --check` and linting as part of pre-push; formatting/linting passed. 
- Ran unit tests for the analysis module with `cargo test -p everruns-server domains::agents::analysis --lib`; all tests passed (11 passed, 0 failed). 
- Ran repository `just pre-push` which exercised the toolchain and checks; it exited non-zero because the environment cannot resolve `origin/main` for the migration immutability check (remote not available), not because of the code changes themselves.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3051060bb0832b91825a1774d75d4d)


chaliy:
Thanks for the review. Dispositions on the threads:

- **Mutex poison → persistent DoS** (`analysis.rs`): fixed — the rate-limiter lock now recovers via `unwrap_or_else(|p| p.into_inner())` instead of `expect()`, so a prior panic can't permanently wedge analysis.
- **Oversized input → 500 vs 4xx**: acknowledged but deferred. The size guard's primary purpose (blocking the paid utility-LLM call) works; distinguishing this to a 4xx cleanly needs a typed error rather than string-matching, which is disproportionate for a >128 KiB edge case. Tracking as a follow-up.
- **`retry_after_seconds`/`code` dropped in `app_a2a` / `app_webhooks` RateLimited arms**: those endpoints don't route `analyze_agent`, and the canonical `From<CommandError>` HTTP adapter (used on the analyze path) preserves the extensions, so 429 callers get the hint where it matters.
- **gRPC `KIND_RATE_LIMITED`**: the gRPC status path already maps `RateLimited` → `resource_exhausted` (8). Adding a proto envelope variant + retry fields is a cross-crate protocol change; deferring as a follow-up.
- **Permit/window ordering and org-deque rollback**: best-effort counters with small self-healing windows; the slight conservatism on rejection is acceptable.

Substantive fix pushed; the rest are minor and tracked.

---

PR #2394, opened 2026-06-21
feat(schedules): min cron interval + per-org cap on session schedules

chaliy:
## What
Enforce a minimum recurring cron interval and a per-org cap on **session schedules**, bringing them to parity with the app `schedule` channel limits shipped in EVE-507. Resolves EVE-614.

## Why
Each session-schedule fire dispatches a real worker turn. Before this change the session-schedule surface enforced only `MAX_ACTIVE_SCHEDULES_PER_SESSION = 5` with:
- **no minimum cron interval** — `* * * * *` (every minute) was accepted, and
- **no per-org cap** — only a per-session count, so unlimited sessions implied unlimited active schedules per org.

On an open-signup deployment that is a sustained amplifier of operator worker/DB compute (independent of who pays for LLM tokens).

## How
Both limits are enforced **worker-side at create time** in the two schedule entry points — the `create_schedule` tool (`crates/core/src/capabilities/session_schedule.rs`) and `spawn_background` with a `schedule` arg (`crates/core/src/tools.rs`) — because session schedules are created on the worker, not via a server command.

- **Min cron interval**: `SESSION_SCHEDULE_MIN_INTERVAL_SECONDS` (default 300s). Recurring crons that fire more frequently are rejected; one-shot (`scheduled_at`) schedules are unaffected. Helper normalizes 5-field and 6/7-field crons and computes the minimum gap; unparseable expressions pass the gate and are rejected later at next-trigger computation.
- **Per-org cap**: `RESOURCE_LIMIT_MAX_SESSION_SCHEDULES_PER_ORG` (default 100), using the `RESOURCE_LIMIT_*` env family so the SaaS wrapper sets it per plan via `PlanResourceLimits::apply_to_env`. Also carried on `ResourceLimitsConfig.max_session_schedules_per_org` for discoverability alongside its siblings.
- New org-scoped count: `count_active_org_schedules` on the `SessionScheduleStore` trait, implemented across DB (Postgres + in-memory), local SQLite, and gRPC stores, backed by a new `CountActiveOrgSchedules` RPC mirroring the existing `CountActiveSessionSchedules`.

Docs: `specs/tool-execution.md` (session schedule limits) and `specs/threat-model.md` (extended TM-SCHED-001).

## Risk
- **Low.** Adds reject-at-create gates and a read-only count path; no migration, no change to how existing schedules fire. Defaults are generous (300s / 100 per org) so legitimate usage is unaffected.
- Possible breakage: an agent attempting a sub-5-minute recurring schedule or an org already past 100 active schedules now gets a clear tool error instead of silently creating one.

## Security
- Threat categories reviewed: **TM-DOS** (this is the mitigation — extended TM-SCHED-001), **TM-TENANT** (per-org count is `org_id`-scoped; the new RPC carries `org_id` set server-side from the resolved session org, identical to the existing schedule RPCs — no new client-controlled tenant boundary), **TM-SQL** (count query is parameterized: `WHERE org_id = $1 AND enabled = true`).
- Findings and resolutions: none. No auth/authz, prompt, or data-exposure surface touched.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

https://claude.ai/code/session_01GDw8evR5xi3yWX3AiX66Ao

---
_Generated by [Claude Code](https://claude.ai/code/session_01GDw8evR5xi3yWX3AiX66Ao)_


chaliy:
Thanks @copilot — addressed all 7 in 88eca74:

- **Duplicated enforcement (tools.rs:1167):** extracted `enforce_create_limits()` (per-session + per-org + min cron interval) into `session_schedule`, now called by both `create_schedule` and `spawn_background`. Preserves the store-error vs limit-rejection distinction via a `ScheduleLimitError` enum.
- **Env-var save/restore in tests (5 comments):** added a test-only `EnvVarGuard` that snapshots and restores the var (value *or* absence) on drop; all schedule limit tests now use it, so they no longer leak `SESSION_SCHEDULE_MIN_INTERVAL_SECONDS` / `RESOURCE_LIMIT_MAX_SESSION_SCHEDULES_PER_ORG` and the misleading "never mutates global env" comment is gone.
- **Doc comment (session_schedule.rs:61):** reworded — this helper is intentionally more permissive than the app channel's `normalize_cron_expression` (it also accepts the 6-field seconds form), no longer implies identical behavior.

Resolving these threads.

---

PR #2444, opened 2026-06-24
fix(subagents): settle idle turns from terminal events

chaliy:
### Motivation
- Foreground waits returned raw `idle` from `wait_for_idle`, but core only treats explicit terminal strings (e.g. `completed`, `failed`) as terminal, leaving successful child subagents unsettled. 
- The change prevents durable spawn handles and session tasks from remaining `Running` when a child turn actually completed.

### Description
- When `SessionStatus::Idle` is observed in the adapters, query the session event tail and derive a terminal turn status (`completed`, `failed`, `cancelled`, `failed` for sealed) instead of returning raw `idle`. 
- If no terminal turn event is visible yet, continue polling to cover the status/event persistence race so callers can distinguish successful vs failed idle turns. 
- Implemented the helper and behavior in `crates/server/src/direct_worker_adapters.rs` and `crates/worker/src/grpc_adapters.rs`. 
- Preserves the core `terminal_subagent_status` mapping (requires explicit `completed`/`failed`/`cancelled`/`max_iterations_reached`).

### Testing
- Ran formatter check with `cargo fmt --check` and ensured formatting was applied where needed. (passed)
- Ran the focused unit test `cargo test -p everruns-core terminal_subagent_status_maps_only_terminal_wait_states -- --nocapture` and it passed. 
- Built/checked affected crates with `cargo check -p everruns-worker -p everruns-server` and the checks completed successfully. 
- Verified repository hygiene with `git diff --check` and committed the change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a38149d7e64832ba0062f5f00f2bdfc)

### DIFF
--- a/crates/core/src/capabilities/agent_handoff.rs
+++ b/crates/core/src/capabilities/agent_handoff.rs
@@ -30,6 +30,7 @@ fn terminal_handoff_status(wait_status: &str) -> Option<SubagentStatus> {
         "error" | "failed" => Some(SubagentStatus::Failed),
         "cancelled" => Some(SubagentStatus::Cancelled),
         "max_iterations_reached" => Some(SubagentStatus::MaxIterationsReached),
+        "sealed" => Some(SubagentStatus::Sealed),
         _ => None,
     }
 }
@@ -619,9 +620,9 @@ impl Tool for StartAgentHandoffTool {
         ) {
             let task_state = match subagent_status {
                 SubagentStatus::Completed => SessionTaskState::Succeeded,
-                SubagentStatus::Failed | SubagentStatus::MaxIterationsReached => {
-                    SessionTaskState::Failed
-                }
+                SubagentStatus::Failed
+                | SubagentStatus::MaxIterationsReached
+                | SubagentStatus::Sealed => SessionTaskState::Failed,
                 SubagentStatus::Cancelled => SessionTaskState::Canceled,
                 SubagentStatus::Running | SubagentStatus::Spawning => SessionTaskState::Running,
             };
--- a/crates/core/src/capabilities/subagents.rs
+++ b/crates/core/src/capabilities/subagents.rs
@@ -89,6 +89,9 @@ fn terminal_subagent_status(wait_status: &str) -> Option<crate::session::Subagen
         "error" | "failed" => Some(crate::session::SubagentStatus::Failed),
         "cancelled" => Some(crate::session::SubagentStatus::Cancelled),
         "max_iterations_reached" => Some(crate::session::SubagentStatus::MaxIterationsReached),
+        // A sealed turn (no forward progress / budget exhausted) is terminal but
+        // distinct from a failure — surface it so the parent can decide next steps.
+        "sealed" => Some(crate::session::SubagentStatus::Sealed),
         _ => None,
     }
 }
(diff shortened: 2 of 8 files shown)


chaliy:
Deep review — **NEEDS-CHANGES** (not mergeable as-is). CI is red and the change appears to break the cancellation/DLQ idle paths.

**1. CI red — directly-relevant test fails (blocker).** `Build Check` and `Integration Tests (PostgreSQL)` both fail. `crates/server/src/grpc_service/tests.rs:485` (`test_subagent_and_handoff_tools_complete_over_grpc_platform_adapter`) asserts `spawn_value["status"] == "idle"`, but the PR now returns a derived terminal status → `left: "failed"` vs `right: "idle"`. The PR description only lists `cargo fmt --check`, one core unit test, and `cargo check`; this integration test was never run.

**2. Cancellation of a subagent regresses to a ~300s hang.** `crates/worker/src/session_lifecycle.rs` `cancelled()` idles the session by emitting **only** `session.idled` — there is no `turn.cancelled` emission anywhere in `crates/worker`. So under this PR a cancelled foreground subagent observes `Idle`, `latest_terminal_turn_status()` returns `None`, and the loop spins until the 300s `wait_for_idle` timeout (`crates/core/src/capabilities/subagents.rs:636` passes `Some(300)`), after which the spawn handle is never settled. That is worse than the bug being fixed.

**3. DLQ/error idling has the same defect.** `dlq_error()` emits `output.message.completed` + `session.idled` with no terminal turn event → same hang + unsettled handle.

**4. Test fake is unrealistic.** `CompletingTestRunner` (`grpc_service/tests.rs:39-69`) emits only `output.message.completed` and sets idle; it never emits `turn.completed`, so even after fixing the assertion the test wouldn't model the real happy path. No new test exercises `latest_terminal_turn_status()` or the no-terminal-event fall-through.

The happy path is correct (`turn_completed()` emits `turn.completed` before `session.idled`), and the DESC+limit query routing is correct. Security: no concerns.

**Fix options (needs a decision — these differ architecturally):**
- (a) Make the idle→terminal derivation tolerant: when idle is observed with no terminal turn event, settle as `completed` (or add `session.idled` to the queried types). Small, but loses the cancelled/failed distinction on those paths.
- (b) Emit proper `turn.cancelled` / `turn.failed` events from `cancelled()` / `dlq_error()` in the worker so the derivation has a real signal. More correct, larger, and must stay consistent across both `direct_worker_adapters.rs` and `grpc_adapters.rs`.

Either way: fix the `status == "idle"` assertions (lines 485, 540), have the test fake emit `turn.completed`, add a regression test proving the no-terminal-event idle path settles without a 300s hang, and reference the `events::TURN_*` constants instead of hardcoded literals.

---

chaliy:
Rebased onto latest `main` and reworked so sealed turns are surfaced to the parent rather than collapsed into `failed`.

**What changed**
- New first-class `SubagentStatus::Sealed`; adapters map `turn.sealed → "sealed"` (both the direct/server and gRPC/worker paths), and `terminal_subagent_status` / `terminal_handoff_status` recognize it. A sealed child settles as a terminal, non-retryable task, but the parent sees `status: "sealed"` (distinct from `failed`) and can decide what to do next. The seal **reason** (`no_progress`/`budget`) reaches the parent via the child's seal-explanation assistant message, which becomes the spawn `result`.

**Root-cause fix found while verifying end-to-end**
- The idle-settling probe (`list_events`) failed with *"Event service not configured"* over the **gRPC worker topology**, because `domain_ctx_for_caller` (grpc_service) never wired `EventService` into the platform-command `Ctx`. That made `wait_for_idle` error and every subagent/handoff foreground wait return `status: "failed"` — not just sealed ones. Fixed by wiring `EventService` into that `Ctx`.
- The grpc test's completing runner emitted only an assistant message + `idle` with no terminal turn event (unrealistic vs. production, which always emits `turn.completed`). It now emits a `turn.completed`, and the assertions verify the child settles as `completed`.

**Verification:** `cargo test -p everruns-server --lib grpc_service::tests::test_subagent_and_handoff_tools_complete_over_grpc_platform_adapter` passes; `cargo clippy -p everruns-core -p everruns-worker -p everruns-server --all-targets` clean; OpenAPI spec unchanged (`SubagentStatus` isn't in the published schema). `specs/subagents.md` updated.

---

PR #2577, opened 2026-07-07
fix(sessions): reject forged parent links

chaliy:
### Motivation
- `parent_session_id` is internal-only metadata used for subagent/handoff ownership. Untrusted callers must not be able to forge it (cross-session/cross-org parent links), but the trusted spawn path must be able to set it.

### Description (revised)
The original command-layer guard (`reject when parent_session_id.is_some() && !ctx.caller.is_internal`) was incorrect: the trusted subagent/handoff spawn path dispatches `CreateSession` through `DirectPlatformStore` as the **session owner** (`is_internal == false`, see `caller_resolution.rs` — "Do not replace with `Caller::internal`"), so the guard rejected legitimate spawns and failed the PostgreSQL integration test.

Enforcement belongs at the untrusted boundary, where it already lived:
- Extract the existing inline strip in `crates/server/src/api/sessions.rs` into `strip_internal_only_fields`, which drops any client-supplied `parent_session_id` before dispatch. This is the only external HTTP entry point that reaches `CreateSession`; the voice API and gRPC `platform_create_session` hardcode `None`, and MCP `tool_agent_run` never sets it.
- Remove the `is_internal` guard from `CreateSession::execute` so the trusted spawn path keeps working.
- Regression tests: `strip_internal_only_fields_drops_client_supplied_parent_session_id` (boundary) and `create_session_with_non_internal_owner_sets_parent_session_id` (a non-internal owner caller successfully creates a session with a parent link).

### Testing
- `cargo clippy -p everruns-server --all-targets --all-features -- -D warnings` — clean.
- `cargo test -p everruns-server --lib parent_session_id` — both new tests pass.
- Rebased onto latest `main`.

### DIFF
--- a/crates/server/src/api/sessions.rs
+++ b/crates/server/src/api/sessions.rs
@@ -516,17 +516,26 @@ pub async fn create_session(
     Json(req): Json<CreateSessionRequest>,
 ) -> Result<(StatusCode, Json<WithUrls<Session>>), (StatusCode, Json<ErrorResponse>)> {
     check_session_create_rate_limit(&state, org.org_id).await?;
-    // `parent_session_id` is internal-only: the subagent/handoff spawn flow
-    // sets it via the platform store (DirectPlatformStore::create_session
-    // dispatches the command directly). External HTTP callers must never set
-    // it — that would let a client forge cross-session/cross-org parent links —
-    // so strip any client-supplied value at the public boundary.
     let mut req = req;
-    req.parent_session_id = None;
+    strip_internal_only_fields(&mut req);
     let urls = UrlBuilder::from_auth_config(&state.auth.config);
     let session = CreateSession(req).run(&state.ctx(&org)).await?;
 
     Ok((StatusCode::CREATED, Json(urls.wrap(session))))
+}
+
+/// Strip request fields that only trusted internal dispatch paths may set.
+///
+/// `parent_session_id` is internal-only metadata used for subagent/handoff
+/// ownership. The trusted spawn flow sets it via `DirectPlatformStore`, which
+/// dispatches the `CreateSession` command directly and never passes through
+/// this HTTP handler. Any value on an inbound public request is therefore a
+/// forgery attempt (cross-session/cross-org parent link) and is dropped here,
+/// at the untrusted boundary — the command layer intentionally trusts the
+/// caller-set value so the spawn path (which runs as the session owner, not an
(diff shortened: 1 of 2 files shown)


chaliy:
## Review: holding this PR — the guard breaks the internal spawn path

Deep review + CI both flag a real regression, so I'm not merging this as-is.

**Problem:** `CreateSession::execute` now rejects any request carrying `parent_session_id` unless `ctx.caller.is_internal` (`crates/server/src/domains/sessions/commands.rs:74-78`). But the legitimate subagent/handoff spawn path dispatches `create_session` through `DirectPlatformStore::create_session`, whose `Ctx` caller comes from `caller_for_user(...)`, which **hard-sets `is_internal: false`** (`crates/server/src/auth/caller_resolution.rs:33`, with an explicit "Do not replace this with `Caller::internal`" comment). So the guard fires on the trusted spawn path and returns `400`.

**Evidence:** `Integration Tests (PostgreSQL)` fails on this branch. Both `spawn_subagent` (`crates/core/src/capabilities/subagents.rs`) and `agent_handoff` (`crates/core/src/capabilities/agent_handoff.rs`) route through the platform store's `create_session` with a real `parent_session_id`, so both break. The added unit test only covers the reject case with a non-internal caller and uses `Caller::internal` for the "allowed" `test_ctx`, which does **not** match the real spawn caller — so it gives false confidence.

**Suggested direction:** don't gate on `is_internal`. Either (a) strip `parent_session_id` at each *untrusted* command-dispatch boundary (the REST handler already does this at `api/sessions.rs`), or (b) have `DirectPlatformStore` mark its spawn dispatch as trusted for the parent link via a dedicated signal. Any fix should add a regression test that a spawn with a **non-internal owner caller** still succeeds.

The other six aardvark PRs in this batch are being rebased/reviewed/merged; this one needs a revision first.

---

chaliy:
Revised and re-pushed (rebased onto latest `main`). Per the analysis above, the `is_internal` command-layer guard was both wrong (it broke the trusted spawn path — the failing PG integration test) and unnecessary: the security invariant is already enforced at the sole untrusted boundary, which strips `parent_session_id`. This revision extracts that strip into `strip_internal_only_fields`, removes the guard, and adds regression tests for both the boundary strip and a non-internal owner successfully setting a parent link. Clippy + the new tests pass locally; merging once CI is green.

---

PR #2662, opened 2026-07-10
fix(auth): normalize emails case-insensitively

chaliy:
## What changed
User emails are now normalized (trimmed + lowercased) at the storage boundary, and uniqueness is enforced case-insensitively. `Alice@example.com` and `alice@example.com` resolve to a single account for register, login, OAuth linking, password reset, resend, and verify.

Normalization is applied once in `StorageBackend::create_user`, `create_user_with_id`, and `get_user_by_email` — the delegating layer over both the Postgres and in-memory backends — so every writer and every lookup canonicalizes email without touching each route. Migration `099` lowercases existing rows and replaces the case-sensitive `idx_users_email` with `UNIQUE (lower(email))`.

## Why
Storage and lookup were case-sensitive with no normalization, while rate-limit keys and org-invite matching already lowercased — an internally inconsistent identity surface. Case variants produced duplicate accounts for one mailbox, "Invalid email or password" when a user varied casing, and OAuth linking that silently created a duplicate instead of linking on a case mismatch (undermining TM-AUTH-012). Fixes EVE-704.

## Before / After
- **Before:** `register John@x` then `register john@x` → two accounts; `login john@x` after signing up as `John@x` → auth failure.
- **After:** the second register collides on `UNIQUE (lower(email))`; login/reset/OAuth all resolve to the one stored (lowercased) account.

Evidence (local): new `email_normalization_tests` pass — `normalize_email` trims/lowercases, and a round-trip through `StorageBackend::in_memory()` shows a mixed-case create is stored lowercased and is found via `alice@…`, `ALICE@…`, and whitespace-padded variants. `cargo fmt --check` clean, `cargo clippy --all-targets --all-features -D warnings` clean, `scripts/lib/check-migration-ordering.sh` reports `001..099` sequential.

## Risk
- Medium (touches a DB migration on `users`).
- The migration lowercases existing emails and builds `UNIQUE (lower(email))`. If a deployment already has case-variant duplicate accounts for one mailbox, the unique-index build fails by design (documented in the migration) so the operator reconciles them rather than the migration silently merging or dropping data. Fresh databases (CI) have no such rows.

## Security
- Threat categories reviewed: TM-AUTH (TM-AUTH-012 OAuth linking, TM-AUTH-017 identity gate), TM-TENANT (org-invite email matching), TM-SQL (migration).
- Findings and resolutions: restores the "one verified account per mailbox" invariant that TM-AUTH-012's linking-safety reasoning assumed; updated that entry to note storage-boundary normalization and the case-insensitive unique constraint. Migration uses only static DDL/`lower()`; no injection surface. No new dependencies.

## Follow-ups
- No follow-ups. (A data migration that auto-merges pre-existing case-duplicate accounts across all FK references is intentionally out of scope — merging identities is destructive and should be an operator-driven action, not a silent migration; the migration fails loudly instead.)

## Checklist
- [x] Tests added or updated

https://claude.ai/code/session_019zyLRqmSQYsq7AiT8n6nBh

---
_Generated by [Claude Code](https://claude.ai/code/session_019zyLRqmSQYsq7AiT8n6nBh)_


chaliy:
Closing as a duplicate of #2661, which implements the same case-insensitive email normalization for EVE-704 (storage-boundary `normalize_email`, migration to `UNIQUE (lower(email))` that fails loudly on pre-existing case-duplicates) with broader validation. Deferring to #2661.

---

PR #2683, opened 2026-07-11
fix(server): bound message-history event reads for long sessions

chaliy:
## What changed
The per-turn conversation-history read now fetches a compact projection instead of full event rows, and slow reads emit one structured warning.

- **Compact projection.** `list_message_events` / `list_message_events_limited` / `list_message_events_filtered` (PostgreSQL and in-memory backends) now return a `MessageEventRow` with only the columns message reconstruction uses — `id`, `sequence`, `event_type`, `data` — dropping the heavy `context`/`metadata`/`tags` JSONB (and `session_id`/`ts`/`created_at`) from the wire and from deserialization. The set of messages returned is byte-for-byte identical, so model-visible conversation correctness is unchanged.
- **Structured slow-query warning.** When a message-history read crosses `SLOW_QUERY_WARN_MS` (default 500 ms) it emits exactly one `WARN` with a stable `operation`/`query_family`, `elapsed_ms`, `row_count`, `limit`, and host-supplied `environment`/`release` (from `OTEL_ENVIRONMENT` / `OTEL_SERVICE_VERSION`). The API accepts only scalars, so SQL params, event content, and PII cannot leak.

Resolves EVE-730.

## Why
A live dev run hit the SQLx slow-query threshold loading message events for one session: **3,031 rows in 1.12 s**, family `events WHERE session_id=$1 AND event_type IN ('input.message','output.message.completed','tool.completed') ORDER BY sequence ASC LIMIT $2`.

Root cause: the per-turn history is rebuilt via `MessageRetriever::load_filtered`, which for a normal turn carries no limit and, on the worker's `SessionAdapter`, resolves to the unbounded `list_message_events_limited(None)` safety-capped read. Every loaded row is sent to the model, so rows cannot be dropped without changing what the model sees. The fix therefore attacks the **network / deserialization / memory** amplification named in the issue rather than the row count: the query is already index-supported, so per-row bytes are the lever. DB execution time was never the problem — the predicate is served by the partial index `idx_events_messages(session_id, sequence) WHERE event_type IN (...)` (migration 084).

## Before / After
**Rows/bytes per read (3,031-message session):**

| | Before | After |
|---|---|---|
| Rows returned | 3,031 | 3,031 (unchanged — correctness preserved) |
| Columns transferred/deserialized per row | 10 (incl. `context`, `metadata`, `tags` JSONB) | 4 (`id`, `sequence`, `event_type`, `data`) |
| Query count | 1 | 1 |
| DB plan | index range scan on `idx_events_messages` | index range scan on `idx_events_messages` (unchanged) |
| Slow read (≥ threshold) | silent (only SQLx's generic 1 s log) | one structured `WARN` mappable to Sentry |

**Structured warning fields (redacted):** `operation=message_history_read`, `query_family=message_history_read`, `elapsed_ms`, `row_count`, `limit`, `environment`, `release`. No `session_id`, SQL params, or event content.

**Query plan / index (documented, live EXPLAIN deferred — see Follow-ups):** both the DESC-then-ASC tail read and the ASC safety-capped read match the partial index predicate exactly, so PostgreSQL uses `idx_events_messages` for an index range scan bounded by `LIMIT`; the narrowed SELECT only removes columns from the projection, which reduces heap-fetch payload and client-side work without changing the access path.

**Tests (in-memory, drive the real reconstruction + windowing code):**
- `long_session_reconstructs_and_windows_correctly` — builds a 1,010-turn session (3,030 events: user → agent+tool_call → tool_result), asserts full reconstruction count/ordering/roles and tool-call→result pairing, then asserts a windowed `keep_head=3 + limit=30` load keeps the genuine first turn as anchor plus the latest turns with pairing intact across the head/tail boundary.
- `slow_query` unit tests assert the required fields, that a release always resolves, threshold gating, and redaction (no SQL/PII markers surface).

## Risk
- **Low.** Storage-layer projection change; identical message set returned. The only semantic reduction is that the compact rows no longer carry `context`/`metadata`/`tags` — audited: every message-reconstruction consumer (worker turn context, `load_messages`, `get_message`, `platform_get_messages`, API message list, retriever) reads only `id`/`sequence`/`event_type`/`data`. One test that asserted event-row provenance `metadata` was repointed to the full-row `list_events` path.

## Security
- **Threat categories reviewed:** TM-DOS, TM-TENANT, TM-SQL, and log data-exposure.
  - **TM-DOS** — reduces message-read amplification (compact projection) and adds latency observability; added `TM-DOS-030` to `specs/threat-model.md`. Row count still bounded by `MESSAGE_SAFETY_LIMIT`.
  - **TM-TENANT** — queries remain session-scoped (`session_id` still bound as `$1`); `MessageEventRow` carries no cross-session data. Reconstructed `Event.session_id` is filled from the caller's own session argument.
  - **TM-SQL** — the projection is a static string; all filters/limits stay parameterized; no user input reaches the SELECT list.
  - **Data exposure** — the slow-query warning is scalar-only by construction (redaction test enforces it); no SQL params, event content, or PII are logged.
- **Findings:** none requiring code change beyond the above.

## Follow-ups
- **Live `EXPLAIN (ANALYZE, BUFFERS)` + end-to-end timing on a representative long session** — deferred: requires a populated PostgreSQL instance, which is not available in this environment. The plan reasoning and supporting index are documented above and in code comments; validate on CI/staging. Safe to defer because the access path is unchanged and already index-backed — this change only narrows the projection.
- PostgreSQL integration tests (`repository_integration_test.rs`) and worker gRPC paths run in CI (no local Postgres/Valkey/NATS here).

## Checklist
- [x] Tests added or updated
- [x] Backward compatibility considered (internal storage API; no external contract change)

---
_Generated by [Claude Code](https://claude.ai/code/session_01D21yJXXdWU3jhfTo8RpwNs)_


chaliy:
Closing as a duplicate of #2682, which was merged first and resolves EVE-730 more completely (capability-aware filtered loading that reduces the row count itself, plus a live `EXPLAIN (ANALYZE, BUFFERS)` and full-stack benchmark). This PR (compact-projection approach) was developed in parallel and now conflicts with the merged fix. No further action — EVE-730 is resolved by #2682.

---

PR #2762, opened 2026-07-14
fix(drivers): bound streaming reconnect first item wait

chaliy:
### Motivation
- The streaming reconnect helper awaited the first SSE event / byte chunk using the HTTP client's read timeout (300s), which let a provider return HTTP 200 and then sit silent, amplifying retries and allowing a single stalled provider to exhaust worker capacity. 
- The change bounds the setup-time peek to the Reason atom's stall budget so silent-header stalls fail quickly instead of retrying with the longer transport timeout.

### Description
- Add `FIRST_STREAM_ITEM_TIMEOUT` (120s) and introduce timed variants of the reconnect helpers so the pre-first-item peek is wrapped with `tokio::time::timeout` and surfaces an `AgentLoopError::llm` on timeout. 
- Preserve existing reconnect semantics for immediate pre-first-item transport errors (body/decode/connect/request/timeout) while preventing retry amplification for silent 200 responses. 
- Add regression coverage for a silent first SSE event and silent first byte chunk, add a `Silent` scripted server behavior, and make test clients bypass system proxy (`no_proxy`) for localhost. 
- Keep SSE and byte-stream helpers behavior-compatible by replaying the peeked first item and returning `RetryMetadata` as before.

### Testing
- Ran formatting and lint: `cargo fmt --check` (passed). 
- Ran focused unit tests: `cargo test -p everruns-core stream_reconnect -- --nocapture`, which exercised the new `silent` regressions and the existing reconnect cases and reported all tests passing (7 passed). 
- Verified the added tests exercise both SSE and raw byte-stream reconnect helpers and assert that silent-header stalls fail at the bounded setup timeout rather than triggering multi-attempt retries.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52a037a698832baaa9b94081951a2f)


chaliy (commented):
Reviewed and prepared for merge.

**Correctness:** Bounds the setup-time first-item peek (both SSE `connect_sse_with_reconnect` and raw `connect_bytes_with_reconnect`) with `tokio::time::timeout(FIRST_STREAM_ITEM_TIMEOUT = 120s)`, surfacing an `AgentLoopError::llm` stall error on elapse. This closes the amplification vector where a provider returns HTTP 200 then goes silent, previously bounded only by the 300s transport read timeout and re-retried. Fail-fast (llm error, not a reconnectable transport error) is the right call — it matches the Reason atom's provider stall budget rather than the longer read timeout. Public helpers cleanly delegate to `_timeout` variants so tests inject a short bound; real transport-error reconnect semantics and first-item replay are preserved. Both new regressions (silent first SSE event / silent first byte chunk) plus the 5 existing reconnect tests pass locally (`cargo test -p everruns-core --lib stream_reconnect` → 7 passed).

Rebased onto latest `main` (clean). Force-pushed. Waiting on CI + other reviewers before merge.

---

PR #2764, opened 2026-07-14
fix(apps): reject future cron bursts

chaliy:
### Motivation
- Prevent schedule-channel validation from accepting cron expressions that have high-frequency bursts only after the 366-day scan horizon, which allowed future-year one-second bursts to bypass the configured `SCHEDULE_CHANNEL_MIN_INTERVAL_SECONDS` safety limit.

### Description
- Update `cron_min_interval_seconds` to keep the existing 366-day horizon scan but fallback to sampling the next few upcoming occurrences when the horizon has fewer than two occurrences to avoid returning `None` for future-dated schedules.
- Introduce a `FALLBACK_OCCURRENCE_LIMIT` and early-return the computed minimum when an interval below the rejection threshold is observed during sampling.
- Add regression unit tests `cron_min_interval_detects_future_year_burst` and `schedule_channel_rejects_future_year_burst` to verify the helper and the schedule-channel validation now detect and reject the future-year burst case.
- Changes are localized to `crates/server/src/domains/apps/commands.rs` (cron validator and tests).

### Testing
- Ran `cargo fmt --check` and the formatting check passed.
- Ran focused unit tests `cargo test -p everruns-server cron_min_interval --lib` and `cargo test -p everruns-server schedule_channel_rejects_future_year_burst --lib`, and both test runs passed.
- Ran local validation checks including `git diff --check` and observed no failures related to the change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4d6df56f98832b99fbe9bcd1ab0968)


chaliy (commented):
Reviewed, rebased, and fixed a bug found in review.

**Core change (correct):** `cron_min_interval_seconds` keeps the 366-day horizon scan and, when the horizon yields fewer than two occurrences, falls back to sampling the next `FALLBACK_OCCURRENCE_LIMIT` (3) upcoming triggers — so a future-year one-second burst (e.g. `* * * 1 1 * 2029`) can no longer bypass `SCHEDULE_CHANNEL_MIN_INTERVAL_SECONDS` by sitting past the horizon. The two regression tests correctly cover the helper and the schedule-channel validation.

**Bug I found and fixed:** the fallback loop restarts iteration from `start`, but `previous` still held the single horizon occurrence from the first loop. For a schedule with **exactly one** occurrence inside the horizon (e.g. a once-yearly schedule), the fallback's first sampled occurrence equals that same time, so `interval = next - previous = 0` — a spurious sub-limit "burst" that would **falsely reject a valid sparse schedule**. Fixed by resetting `previous = None` before the fallback loop, and added `cron_min_interval_handles_single_horizon_occurrence` (a once-yearly schedule must report its ~1-year cadence, not 0). All 6 focused tests pass: `cargo test -p everruns-server --lib -- cron_min_interval schedule_channel_rejects_future_year_burst` → 6 passed.

**Rebase:** conflicted on the fn signature (main had made it `pub(crate)`); kept `pub(crate)` and the new doc/const. Rebased onto latest `main`. Waiting on CI + other reviewers before merge.

---

PR #2832, opened 2026-07-18
fix(agent-triggers): preserve migrated app schedule context

chaliy:
### Motivation

- Migration 106 retargeted App schedule channels to Agent triggers but dropped the App execution context (harness, owner principal, resolved user, agent identity), which could widen execution capabilities and change user-scoped memory after upgrade.
- Preserve existing App semantics for migrated schedules to avoid a silent security context shift while leaving native Agent triggers unchanged.

### Description

- Add nullable execution context columns to `agent_triggers` (`execution_harness_id`, `execution_owner_principal_id`, `execution_resolved_owner_user_id`, `execution_agent_identity_id`, `execution_app_id`) and update the storage model and repository SQL to persist them.
- Update migration `106_migrate_app_schedules_to_agent_triggers.sql` to copy the App execution fields into migrated Agent triggers instead of only agent/config/schedule fields.
- Change invocation logic in `agent_triggers::commands` to resolve a `TriggerExecutionContext` (new `resolve_trigger_execution_context`) and to use the preserved App context when present (calling `create_from_app`) while keeping native Agent-trigger behavior as the fallback (`create_from_agent_trigger`).
- Add a focused unit test `resolve_trigger_execution_context_preserves_migrated_app_context` and update in-memory storage/test fixtures to include the new fields.

### Testing

- Ran `cargo test -p everruns-server --lib domains::agent_triggers --all-features` and all tests passed (12 passed, 0 failed), including the new regression test.
- Ran `cargo fmt --check` which passed.
- Ran `bash scripts/lib/check-migration-ordering.sh` which reported migrations sequential (001..106).
- A broader `cargo test -p everruns-server domains::agent_triggers --all-features` invocation was started but was intentionally interrupted after the server library compiled because it proceeded to build unrelated integration-test binaries, so it did not complete end-to-end; focused library tests provide the targeted verification described above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5a5a29bbd8832b8815fc2b6119b6b4)


chaliy:
Rebased onto `main` and reworked to satisfy the **Migration Immutability** check, which (correctly) rejected the original approach of editing the already-merged migrations `104_agent_triggers.sql` and `106_migrate_app_schedules_to_agent_triggers.sql`.

Changes vs. the original PR:
- Reverted the in-place edits to `104` and `106`.
- Added the five `execution_*` columns via a new sequential migration `108_agent_trigger_execution_context.sql` (`ALTER TABLE agent_triggers ADD COLUMN …`).
- Kept the code changes: `resolve_trigger_execution_context` uses a populated context verbatim (migrated App schedules) and otherwise falls back to the agent's own identity (native triggers, matching current `main`). Also fixed a rebase-surfaced compile break in `repository_conformance_test` (new `CreateAgentTriggerRow` fields).

⚠️ **Backfill caveat (needs a decision):** because `106` is immutable and it deletes the `app_channels` rows during migration, triggers **already** migrated by `106` retain no link back to their source App and therefore cannot be backfilled by `108`. Those rows keep `execution_* = NULL` and fall back to agent-identity ownership — i.e. the migration-time context preservation this PR originally aimed for is not delivered for pre-existing migrated rows. The new columns + code do preserve context for any future context-aware trigger creation. If retroactively restoring context for already-migrated schedules is required, it needs a separate mechanism (the linkage would have to be reconstructed from `apps`/`durable_schedules`, which is not 1:1). Flagging rather than guessing the intended end-state.

---

PR #2877, opened 2026-07-24
fix(server): bind MCP OAuth grants to provider authority

chaliy:
### Motivation

- Prevent refresh-token exfiltration when an MCP OAuth provider is retargeted so a stored refresh token cannot be sent to a different token endpoint. 
- Ensure stored grants remain bound to the OAuth authority that issued them, or are rotated/invalidated when the authority changes. 

### Description

- Make org-managed OAuth MCP server authority immutable by rejecting URL or auth-mode changes in `McpServerService::update`, returning an error that instructs operators to create a new server instead of retargeting the existing one. 
- Change plugin anchor sync (`sync_plugin_oauth_anchors`) to replace the anchor and create a new provider id when a plugin's MCP server URL changes instead of reusing the old provider id. 
- Add unit tests: `changed_url_rotates_provider_id` in `oauth_anchor.rs` and `update_rejects_oauth_server_retargeting` in `mcp_servers::service` to cover plugin- and managed-server retargeting behavior. 
- Update spec text in `specs/plugins.md` to document that URL changes rotate provider ids and require user reconnection. 

### Testing

- Ran `cargo fmt --check` and `git diff --check`, both succeeded. 
- Added focused unit tests for anchor rotation and update rejection; targeted `cargo test -p everruns-server domains::plugins::oauth_anchor::tests --lib` began building but the full dependency compilation was interrupted before completion in this environment. 
- The new tests are included and expected to pass in CI; a full `just pre-push` / continuous integration run should be performed before merge.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c04ee0be4832b87ec61033ca718f4)

### DIFF
--- a/crates/server/src/domains/mcp_servers/commands.rs
+++ b/crates/server/src/domains/mcp_servers/commands.rs
@@ -252,6 +252,22 @@ pub struct UpdateMcpServerCmd {
     pub req: UpdateMcpServerRequest,
 }
 
+/// OAuth authority is immutable. The MCP server row id doubles as the OAuth
+/// provider id, so changing an OAuth server's URL — or toggling it out of OAuth
+/// mode — would let a stored refresh token be replayed against a newly
+/// discovered token endpoint. Returns true when an update tries to retarget an
+/// OAuth server's authority and must therefore be rejected.
+fn oauth_authority_retargeted(
+    existing_auth_mode: &McpServerAuthMode,
+    existing_url: &str,
+    req_url: Option<&str>,
+    req_auth_mode: Option<&McpServerAuthMode>,
+) -> bool {
+    *existing_auth_mode == McpServerAuthMode::OAuth
+        && (req_url.is_some_and(|url| url != existing_url)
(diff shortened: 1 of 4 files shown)


chaliy:
Rebased onto latest `main` and closed a gap in the fix during review: the OAuth-authority immutability guard was only on `McpServerService::update`, which has **no non-test callers**. The live `PATCH /v1/mcp-servers/{id}` route dispatches `UpdateMcpServerCmd::execute`, which still applied `url`/`auth_mode` changes unconditionally — so an org-managed OAuth server could still be retargeted, letting a stored refresh token flow to a newly discovered token endpoint (the exact threat this PR targets).

Added a pure `oauth_authority_retargeted(...)` predicate and applied it in `UpdateMcpServerCmd::execute` (the real path), plus 4 unit tests for it. Locally `cargo test -p everruns-server --lib oauth_authority` passes (4/4). The plugin-anchor rotation half of the original change was already correct and is unchanged.

---

PR #2880, opened 2026-07-24
fix(evals): redact structured dataset secrets

chaliy:
### Motivation
- The dataset exporter promised always-on secret scrubbing but only inspected string leaves, allowing structured fields like `{"password":"hunter2"}` to leak secret values.
- Exported `messages` include tool arguments/results and metadata where secrets commonly appear as JSON fields, so sanitization must be key-aware to meet the confidentiality guarantee.

### Description
- Add a `SECRET_KEYS` list and `is_secret_key` helper that matches sensitive JSON keys case-insensitively and ignores `-`/`_` differences (covers `api_key`, `access_key`, `secret`, `password`, `passwd`, `token`, etc.).
- Update `sanitize_value` to redact the value (replace with `REDACTED`) when a map key matches a secret key while preserving existing `redact_content` behavior for content-bearing fields.
- Keep existing string-pattern scrubbing (`scrub_secrets`) for token-like substrings and preserve `SECRET_PATTERNS` as an additional defense.
- Add a regression test `sanitize_value_redacts_structured_secret_fields` that verifies nested structured secret fields are redacted while safe sibling fields remain intact.

### Testing
- `cargo fmt --check` passed locally against the modified file.
- `git diff --check` passed with no whitespace or diff errors.
- Attempted the targeted unit test `cargo test -p everruns-server domains::evals::dataset::tests::sanitize_value_redacts_structured_secret_fields`, but the cold Rust build stalled after compilation/linking and was stopped in this environment.
- Attempted `timeout 180 cargo check -p everruns-server`, which timed out while compiling dependencies in this environment, so a full CI run is required to validate the change end-to-end.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5baec9cde8832b82f00088dc577f03)

### DIFF
--- a/crates/server/src/domains/evals/dataset.rs
+++ b/crates/server/src/domains/evals/dataset.rs
@@ -81,6 +81,24 @@ const CONTENT_KEYS: [&str; 7] = [
     "base64",
 ];
 
+/// Structured JSON keys whose values are credentials even when the value itself
+/// does not match a standalone token pattern.
+/// Substrings (matched against the normalized, separator-stripped, lowercased
+/// key) that mark a JSON field as carrying a credential. `contains` matching is
+/// used so compound names like `openai_api_key`, `client_secret`, or `x-api-key`
+/// are covered, not just the bare words.
+const SECRET_KEY_SUBSTRINGS: [&str; 9] = [
+    "apikey",
+    "accesskey",
+    "secretkey",
+    "privatekey",
+    "secret",
+    "password",
+    "passwd",
+    "credential",
+    "authorization",
+];
+
 /// High-signal credential patterns scrubbed from every exported string, always.
 /// Deliberately conservative to avoid mangling legitimate content.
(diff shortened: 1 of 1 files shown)


chaliy:
Rebased onto latest `main` and hardened the redaction during review. The original `is_secret_key` used exact-match on the normalized key, so common compound credential names still leaked: `openai_api_key` → `openaiapikey` ≠ `apikey`, plus `client_secret`, `x-api-key`, `refresh_token`, `authorization`, `private_key`, etc.

Switched to substring matching over a curated keyword list, with a `*token`-suffix rule so bearer/refresh/access/id tokens are redacted while token **count** fields (`input_tokens`/`output_tokens`, plural) are preserved and not corrupted. Extended the regression test to cover the compound keys and the count-preservation case. Locally `cargo test -p everruns-server --lib sanitize_value` passes.

---

PR #2920, opened 2026-08-03
fix(sessions): mount registry skills referenced as skill capabilities

shbodya:
## What changed
Registry skills attached to harnesses/agents/sessions as `skill:{uuid}` capability refs are now actually mounted into the session VFS under `/.agents/skills/{name}/` (reconstructed `SKILL.md` + bundled text files), where the built-in `SkillsCapability` discovers them. `list_skills` and `<available_skills>` now include them. Covers both session creation and session repair (both flow through `collect_capability_mounts`).

## Why
The feature was a silent no-op: `skill:{uuid}` refs validated, appeared in the capability catalog, and were accepted on harnesses and agents — but dependency resolution passes only `declarative:` and `plugin:` refs through when a capability has no registry entry, so `skill:` refs were dropped before mount collection. `AttachSkillCapability` existed and was fully tested but was referenced only from tests. Sessions received `/memory/*` mounts from the same pass but never `/.agents/skills/*`.

## Before / After
Before: harness carries `skill:{uuid}` → session VFS has no `/.agents/skills/` entries, `list_skills` returns empty, agents report the skill missing.
After: session VFS contains `/.agents/skills/{name}/SKILL.md` (+ bundled files); integration test drives catalog → agent attach → session create → reads the mounted SKILL.md through the session fs API.

## Risk
- Low / Medium
- New behavior only for refs that previously did nothing. Failure mode choice: a missing or non-active referenced skill now fails session creation loudly (mirrors workspace-memory mounts) instead of degrading silently — an org that archived a still-referenced skill will see session creation errors until the ref is detached or the skill restored.

## Security
- Threat categories reviewed: TM-API (org scoping).
- Findings and resolutions: skills are loaded via `get_skill(org_id, uuid)` — org-scoped, consistent with `validate_capability_refs`; mounts are read-only. No cross-org access path added.

## Follow-ups
- Binary bundled skill files are skipped with a warning (`AttachSkillCapability` file API is text-only); wire base64 `MountEntry` support to mount them.
- Dependency resolution still drops `skill:` refs from `resolved_ids`; mounts are collected from the raw config list instead. Passing them through the resolver (like `declarative:`/`plugin:`) would be the deeper unification.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/crates/server/src/domains/sessions/service.rs
+++ b/crates/server/src/domains/sessions/service.rs
@@ -33,13 +33,13 @@ use everruns_core::{
     PrincipalSummary, Rule, Session, SessionFile, SessionId, SessionSeedMode, SessionStatus,
     TokenUsage, WorkspaceId,
     capabilities::{
-        MEMORY_CAPABILITY_ID, RiskLevel, SystemPromptContext, collect_capabilities_with_configs,
-        compute_features, resolve_capability_configs,
+        AttachSkillCapability, MEMORY_CAPABILITY_ID, RiskLevel, SystemPromptContext,
+        collect_capabilities_with_configs, compute_features, resolve_capability_configs,
     },
     is_declarative_capability, is_mcp_capability, is_plugin_capability, is_skill_capability,
     memory::{MemoryConfig, MemoryMountAccess},
     merge_capabilities, merge_initial_files, normalize_initial_file_path,
-    parse_declarative_capability_id,
+    parse_declarative_capability_id, parse_skill_capability_id,
     typed_id::MemoryId,
 };
 use everruns_durable::UpdateField;
@@ -1107,6 +1107,14 @@ impl SessionService {
             self.collect_workspace_memory_mounts(org_id, &resolved_configs)
                 .await?,
         );
+        // `skill:{uuid}` refs have no registry entry, so dependency resolution
+        // drops them from `resolved_configs` — resolve them from org data
+        // against the raw config list instead (same pattern as declarative and
+        // plugin refs, which carry their definition in the config payload).
+        mounts.extend(
+            self.collect_registry_skill_mounts(org_id, &capability_configs)
+                .await?,
+        );
         ensure_no_reserved_memory_mounts(&mounts)?;
         if let Some(scoped_memory) = scoped_memory {
             mounts.extend(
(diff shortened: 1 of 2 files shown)


shbodya:
Note: #2921 also appends tests to `crates/server/tests/skills_integration_test.rs`; whichever of the two merges second will need a trivial rebase (both hunks append at end of file).

crates/server/src/domains/sessions/service.rs
```diff
@@ -2163,6 +2171,73 @@ impl SessionService {
...
+            })?;
+            if !seen.insert(skill_uuid) {
+                continue;
+            }
+            let row = self
+                .db
+                .get_skill(org_id, skill_uuid)
+                .await?
+                .filter(|row| row.status == "active")
+                .ok_or_else(|| ResourceNotFoundError::new("Skill"))?;
```
chaliy, line 2203: Hard-failing on non-`active` skills makes archiving or disabling a skill break every agent that references it.

`validate_capability_refs` accepts a `skill:{uuid}` ref at **any** status — unlike declarative capabilities, which it restricts to `active`/`disabled`:

https://github.com/everruns/everruns/blob/main/crates/server/src/domains/capabilities/validation.rs#L43-L49

So the attach-time check and this mount-time check disagree. The sequence that bites:

1. Attach a skill to an agent — accepted.
2. Archive (or disable) that skill — accepted, no warning that agents reference it.
3. Every subsequent session creation for that agent fails with a bare 404 `Skill`, with nothing naming which skill or why.

Step 3 is the problem: a soft-delete on one resource silently takes down an unrelated agent, and the error gives the operator nothing to act on. The workspace-memory precedent cited in the doc comment isn't quite parallel — a missing memory mount means the session has no memory to work with, whereas a session missing one skill is still perfectly runnable.

Two options, in order of preference:

1. **Skip non-active skills with a `tracing::warn!`** and mount the rest. Degrades the way a disabled capability should: the session runs, the skill just isn't there.
2. **Keep it fatal but make it diagnosable** — `BadRequestError` naming the skill and its status, e.g. `Skill '{name}' is {status} and cannot be mounted`, rather than `ResourceNotFoundError::new("Skill")`.

Either way, `"disabled"` deserves a separate decision from `"archived"`: it's an accepted state for declarative refs, and it's a strange reason to refuse to start a session.

---

chaliy (commented):
The gap is real and the fix is in the right place — nothing else in `sessions/service.rs` mounted `skill:{uuid}` refs, so those capabilities validated, showed up in the catalog, and mounted nothing. Path traversal isn't a concern here either: skill names go through `validate_skill_name` (lowercase/digits/hyphen only) and archive file paths are traversal-checked at ingest (`skills/archive.rs:160`).

One behavioral issue inline, plus two minor notes:

- **Test doesn't cover the reported symptom.** It asserts the file lands in the VFS, but the bug was `list_skills` coming back empty. `AttachSkillCapability::dependencies()` is `["session_file_system"]`, and this path bypasses dependency resolution entirely — nothing guarantees `SkillsCapability` is in the session to discover the mount. Asserting discovery (or `list_skills`) would pin the actual regression.
- **`file.path != "SKILL.md"` filter is dead.** `extract_zip_archive` consumes `SKILL.md` (and `*/SKILL.md`) before persisting file rows, so no such row exists. Harmless, but the comment claims it drops an "archived copy" that can't be there.

---

crates/server/src/domains/sessions/service.rs
chaliy: Agreed, and taken as written — option 1.

`collect_registry_skill_mounts` now skips a missing or non-`active` skill with a `tracing::warn!` carrying the skill id and status, and the session starts without it. `"disabled"` and `"archived"` deliberately land on the same behavior: neither is a reason to refuse to start a session, and neither is silent anymore.

```
WARN everruns_server::domains::sessions::service: Referenced skill is not active;
     skipping session mount skill_id=019fd488-7e8a-7031-9c3e-0fcf3b36ef9c status=archived
```

Regression test `test_archived_skill_capability_skips_mount_without_failing_session` drives the sequence you described — attach, archive, create session — and asserts 201 on session creation with a 404 on the mount path. Also documented in `specs/skills-registry.md` so the attach-time/mount-time contract is written down.

This branch is on a fork, so the fix went to #2927, which supersedes this PR: same change rebased onto latest `main` (this one had conflicted with #2921 in `skills_integration_test.rs`).


---

chaliy: Lol

chaliy:
Superseded — this landed on `main` in f8e0c8a (#2927), with credit carried over as a `Co-Authored-By` trailer.

What shipped, versus this PR:

- `SessionService::collect_registry_skill_mounts` and the `skills_integration_test.rs` coverage are on `main` essentially as written here.
- One deliberate behavior change: a referenced skill that is missing or not active is **skipped with a warning** instead of failing session creation. `validate_capability_refs` accepts a `skill:{uuid}` ref at any status, so archiving or disabling a skill must not take down every agent referencing it — unlike a missing memory mount, a session missing one skill is still runnable.
- `specs/skills-registry.md` was updated to document the mount path.

The follow-ups noted here are still open and unchanged on `main`: binary bundled skill files are skipped with a warning (needs base64 `MountEntry` support), and `skill:` refs are still collected from the raw config list rather than passed through dependency resolution.

This branch is now conflicted against `main` and has nothing left to add, so closing.

---

PR #2921, opened 2026-08-03
fix(skills): allow restoring archived skills via status-only update

shbodya:
## What changed
Archived skills can be restored: `PATCH /v1/skills/{id}` with a status-only body (`{"status": "active"}` or `"disabled"`) now succeeds on an archived skill and clears `archived_at` in both storage backends. Content edits (`skill_md`) on archived skills remain rejected — with a clearer error telling the caller to restore first — and deleted skills remain immutable. `PATCH` with `status: "deleted"` is still forbidden (destroy stays a separate permission).

## Why
`DELETE /v1/skills/{id}` documents "Archive a skill (soft delete). Can be restored." — but PATCH rejected *any* body on an archived skill with "Archived or deleted skills cannot be edited", including a restore. Archiving was a one-way door that also permanently reserved the skill's unique org-scoped name.

## Before / After
Before: archive a skill → `PATCH {"status":"active"}` → 400 "Archived or deleted skills cannot be edited".
After: same PATCH → 200, skill is `active` with `archived_at: null`, and content edits work again. Integration test covers the full cycle: create → archive → content-edit rejected → status-only restore → content-edit succeeds.

## Risk
- Low
- Loosens one guard along the documented direction only; delete/destroy semantics untouched.

## Security
- Threat categories reviewed: TM-API (permission boundaries).
- Findings and resolutions: restore runs under the same `SKILL_MANAGE` policy as archive; the deleted state and the dangerous-delete permission gate are unchanged.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/crates/server/src/domains/skills/commands.rs
+++ b/crates/server/src/domains/skills/commands.rs
@@ -353,17 +353,30 @@ impl Command for UpdateSkillCmd {
             ));
         }
 
-        // Check existing status
+        // Check existing status. Archiving is documented as restorable
+        // (see DeleteSkill), so archived skills accept a status-only PATCH
+        // (the restore path) but no content edits; deleted skills stay
+        // immutable.
         if let Some(existing) = ctx
             .db
             .get_skill(ctx.org_id(), id)
             .await
             .map_err(classify_anyhow)?
-            && !matches!(existing.status.as_str(), "active" | "disabled")
         {
-            return Err(CommandError::bad_request(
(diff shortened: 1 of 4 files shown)


shbodya:
Note: #2920 also appends tests to `crates/server/tests/skills_integration_test.rs`; whichever of the two merges second will need a trivial rebase (both hunks append at end of file).

PR #2958, opened 2026-08-07
fix(ui): show resolved default chat model

chaliy:
## What changed
The chat model selector now names the exact server-resolved model while the user keeps the `Default` choice, for example `Default · Claude Opus 5`. The label covers loading and unavailable states, preserves explicit-model labels, and refreshes after an organization default-model update.

A new read-only session endpoint exposes the runtime's canonical resolved model ID, so the client does not duplicate model precedence rules. The OpenAPI contract and generated UI types are updated with it.

## Why
Showing only `Default` made it impossible to tell which model was active and could leave users unaware when the effective organization default changed.

## Before / After
Before: the chat selector displayed only `Default`.

After: the selector displayed `Default · Claude Opus 5`; after changing the organization default, the endpoint returned GPT-5.4's model ID and the selector updated to `Default · GPT-5.4` on reload. UI screenshots are attached in a PR comment.

Validation:
- `just pre-push` — all 10 checks passed
- Focused UI tests — 29 passed across model selection, menu, chat panel, and organization invalidation suites
- Focused Rust resolution test — 1 passed
- UI typecheck, lint, and format check — passed
- Live smoke test on `PORT_PREFIX=286` — passed

## Risk
- Low
- The new endpoint adds one small read per mounted chat session; failure degrades the label to an explicit unavailable state without changing turn execution.

## Security
- Threat categories reviewed: TM-AUTHZ, TM-TENANT, TM-API, TM-WEB
- Findings and resolutions: the endpoint reuses the existing `GetSession` authorization policy and organization-scoped storage reads; it returns only a model ID. Errors use the standard sanitized API response path, and labels render as React text. No secrets, new dependencies, or unbounded inputs/work were introduced. No new threat-model entry is needed.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated


chaliy:
## UI smoke evidence

Default resolves to Claude Opus 5:

Default selector resolved to Claude Opus 5

After the organization default changes, the same Default choice resolves to GPT-5.4:

Default selector updated to GPT-5.4

PR #2960, opened 2026-08-07
fix(chat): summarize repeated tool activity

chaliy:
## What changed
Grouped chat activity headlines now collapse repeated equivalent tool actions into localized, count-aware narration such as “Searched files twice” or “Searched files 3 times”. Mixed groups retain the first two distinct action summaries and report the remaining action count, keeping headers bounded while child rows preserve their detailed narration.

## Why
Repeated tool batches produced noisy headers such as “Ran Grep Files and Ran Grep Files” even though the child rows already explained each search. The shared grouping rule now applies to all tool narration rather than special-casing Grep Files.

## Before / After
Before: two Grep Files calls render “Ran Grep Files and Ran Grep Files”.

After: the same collapsed and expanded group renders “Searched files twice”; expanded child rows still render “Searched files for full_name” and “Searched files for login”.

See the attached collapsed and expanded before/after screenshots in the PR conversation.

## Risk
- Low
- Risk is limited to activity-group headline wording and localization. Existing child narration and tool execution are unchanged. Focused unit/integration coverage exercises single, repeated, localized, mixed, and bounded summaries.

## Security
- Threat categories reviewed: TM-TOOL, TM-WEB, TM-DOS.
- Findings and resolutions: No execution, authorization, or trust-boundary changes. Group headlines strip detail arguments while retaining only operation/action discriminators, React text escaping is unchanged, and aggregation is linear over the existing batch with output bounded to two summaries plus a count. No new threat or threat-model update is required.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/crates/core/src/atoms/act.rs
+++ b/crates/core/src/atoms/act.rs
@@ -49,8 +49,8 @@ use crate::tool_fingerprint::{
     tool_call_fingerprint, tool_error_fingerprint, tool_result_fingerprint,
 };
 use crate::tool_narration::{
-    ToolNarrationContext, ToolNarrationPhase, render_group_headline_with_locale,
-    render_tool_narration_with_locale,
+    GroupHeadlineAction, ToolNarrationContext, ToolNarrationPhase,
+    render_tool_narration_with_locale, summarize_group_actions, tool_call_for_group_summary,
 };
 use crate::tool_types::{SideEffectClass, ToolCall, ToolDefinition, ToolResult};
 use crate::traits::{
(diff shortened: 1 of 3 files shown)


chaliy:
### Before (collapsed)

Before: repeated Grep Files activity group collapsed

### Before (expanded)

Before: repeated Grep Files activity group expanded

### After (collapsed)

After: count-aware repeated activity group collapsed

### After (expanded)

After: count-aware repeated activity group expanded with detailed child rows

PR #2964, opened 2026-08-07
fix(plugins): install plugins from larger repositories

chaliy:
## What changed
Plugins hosted inside larger GitHub marketplace repositories can now install successfully while archive extraction remains bounded. The Plugins catalog also renders installation failures inline instead of surfacing an unhandled mutation rejection through the Next.js runtime overlay.

## Why
The default marketplace downloads a pinned whole-repository GitHub tarball before selecting the plugin subdirectory. The current repository archive is about 83.2 MiB unpacked with tar framing, while the Resend plugin itself is only 4.1 KiB, so the former 64 MiB whole-archive bound rejected a legitimate install with `tarball unpacked size exceeds 67108864 bytes (extraction limit)`.

## Before / After
Before: a focused regression reproduces the exact 67,108,864-byte extraction error for a 65 MiB repository even when the requested plugin is small. Rejected catalog mutations could escape as unhandled Promise rejections.

After: the whole-archive decompression bound is 128 MiB, entry iteration is independently capped at 100,000 before subdirectory filtering, and the existing compressed-download, plugin file/count/retained-size, link, and path-containment protections remain. The real default marketplace synced at pinned SHA `d4f07d34d192a1d12a788359fdecb7891002eb72`; installing Resend returned HTTP 201 with active capability `plugin:resend`, version 0.1.0, and no warnings. Manual Chromium verification confirmed the active Resend card and the catalog Installed state. UI screenshots were captured for review evidence.

Validation:
- `just pre-push`
- `cargo test -p everruns-server --lib domains::plugins::fetcher::tests` (9 passed)
- focused Plugins page Jest test
- UI format, lint, typecheck, and production build
- real marketplace sync, Resend install, and Plugins page smoke test with `PORT_PREFIX=273 just start-dev --no-watch`

## Risk
- Medium
- The whole-repository decompression allowance is larger, increasing bounded CPU work for an archive that passes the compressed download cap. The new independent entry cap and existing plugin-specific limits constrain retained memory and iteration. Legitimate repositories above 128 MiB unpacked remain rejected.

## Security
- Threat categories reviewed: TM-DOS, TM-FS, TM-WEB, TM-API, TM-PLUGIN
- Findings and resolutions: the former counter applied only to regular files, so zero-length non-file entries were not charged. Extraction now counts every entry before type/subdirectory filtering and enforces both byte and entry limits. Symlink/hardlink dropping, path traversal/containment checks, compressed download limits, and plugin-specific size/count limits remain in place. React renders API error text without unsafe HTML. `TM-PLUGIN-004` is updated to match the mitigation.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated


chaliy:
### Manual UI verification

Installed Resend plugin card after the real marketplace install:

Resend plugin installed and active

Catalog state after installation:

clipboard

PR #2967, opened 2026-08-07
fix(ui): redact secret store trajectory values

chaliy:
## What changed
Trajectory tool details now receive the tool name when formatting inputs and outputs. For `secret_store`, every structured `value` field is redacted recursively while operation, name, and status metadata remain visible.

A component regression test covers both a `set` input and a `get` result and verifies neither secret reaches the rendered DOM.

## Why
The trajectory path applied only generic key-name redaction. Because `secret_store` carries secret material under the contextual key `value`, session viewers could expand a tool node and read stored secret values.

## Before / After
Before: a rendered `secret_store` tool node contained `input-secret` in Input and `output-secret` in Result.

After: `pnpm exec jest --runInBand src/__tests__/trajectory-nodes.test.tsx` proves both fields render as `[redacted]` and neither sentinel secret is present in the DOM. A screenshot was not captured because the layout is unchanged and reproducing the vulnerable content visually would create an unnecessary secret-like artifact; the rendered-component assertion is the direct evidence.

## Risk
- Low
- The change is limited to expandable trajectory detail formatting. Non-`secret_store` tool details retain existing redaction, while `secret_store` metadata remains visible except for fields named `value`.

## Security
- Threat categories reviewed: TM-WEB, TM-TOOL, TM-TENANT
- Findings and resolutions: confirmed contextual secret values could reach trajectory DOM text; resolved by passing tool identity into formatting and recursively masking `secret_store.value` fields for both arguments and results. No new input, HTML, authorization, dependency, persistence, or unbounded-work surface was introduced. The existing threat model already covers frontend secret disclosure, so no threat-model update is required.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/apps/ui/src/__tests__/trajectory-nodes.test.tsx
+++ b/apps/ui/src/__tests__/trajectory-nodes.test.tsx
@@ -132,4 +132,49 @@ describe("ToolGroupNode", () => {
     expect(screen.getByText(/OPENAI_API_KEY=\[redacted\]/)).toBeInTheDocument();
     expect(screen.queryByText(/sk-live-secret/)).not.toBeInTheDocument();
   });
+
+  it("redacts secret_store values from inputs and results", () => {
+    render(
+      <ToolGroupNode
+        id="tools"
+        type="toolGroup"
+        selected={false}
+        dragging={false}
+        draggable={false}
+        selectable={false}
+        deletable={false}
+        zIndex={0}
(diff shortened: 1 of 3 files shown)


chaliy:
Closing as a duplicate of #2982 (byte-identical diff), which has been reviewed, rebased, passed CI, and merged.

---

PR #2968, opened 2026-08-07
fix(ui): make model rows responsive

chaliy:
## What changed
Model rows now wrap their identity, capability badges, health status, and actions according to the available content width. Shared two-column pages keep their context rail stacked until the wider breakpoint, and Models page headings, organization settings, and expanded profile details also adapt cleanly at compact widths.

A browser regression covers 1024×900 and 768×900 layouts, including action containment, rail stacking, and horizontal overflow. A matching manual test case documents the responsive acceptance criteria.

## Why
At compact desktop widths, the shared page rail consumed 320px while model rows retained a desktop-only horizontal layout. Model names wrapped into narrow columns and actions rendered outside their bordered rows, making the page clipped and difficult to use.

## Before / After
Before: at the reported compact width, model names collapsed to a few words per line and capability/action controls were clipped past the right edge of every row.

After: real-stack Chromium smoke tests at 1024×900 and 768×900 show full-width model rows, a stacked provider rail, visible actions, readable expanded profile details, and `documentElement.scrollWidth === clientWidth`. Screenshots are attached in the PR conversation.

Automated reproduction: the new Playwright test failed before the fix with the Disable button ending at x=865.84 while its row ended at x=656 (1024px viewport). It now passes at both tested widths.

## Risk
- Low
- The shared page rail now stacks between 1024px and 1279px instead of forcing a narrow two-column layout. This is an intentional readability tradeoff across pages using the shared skeleton.
- No data, API, or interaction semantics changed.

## Security
- Threat categories reviewed: TM-WEB
- Findings and resolutions: Presentational layout and accessible-label changes only. Model/provider text remains React-escaped; no raw HTML, URLs, request handling, state mutation, or trust boundary changed. No findings and no threat-model update required.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/apps/ui/e2e/models-responsive.spec.ts
+++ b/apps/ui/e2e/models-responsive.spec.ts
@@ -1,0 +1,140 @@
+import { expect, test, type Page } from "@playwright/test";
+
+const DEFAULT_ORG_ID = "org_00000000000000000000000000000001";
+
+const providers = [
+  {
+    id: "provider-openai",
+    name: "OpenAI",
+    provider_type: "openai",
+    status: "active",
+    api_key_set: true,
+    base_url: null,
+    created_at: "2026-07-01T00:00:00Z",
+    updated_at: "2026-07-01T00:00:00Z",
+  },
+];
+
+const models = [
+  {
+    id: "model-gpt-5-6-luna",
+    model_id: "gpt-5.6-luna",
+    display_name: "GPT-5.6 Luna",
+    provider_id: "provider-openai",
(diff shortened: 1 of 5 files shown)


chaliy:
### Responsive layout evidence

Before (reported compact layout), after at 1024×900, and after at 768×900 with profile details expanded:
clipboard


After — 1024 × 900:
clipboard


After — 768 × 900, expanded profile:
clipboard

PR #2991, opened 2026-08-07
fix(session-files): exclude private memory before grep

chaliy:
## What changed
Non-owner session-file grep now excludes the private `/memory/user` subtree before storage matching, file reads, or scan-budget accounting. The exclusion is propagated through PostgreSQL, blob-backed, in-memory, session-store, and virtual-mount paths, with response redaction retained as defense in depth.

During the rebase, the exclusion was reconciled with main's shared glob matcher: storage returns path-filtered metadata candidates, private paths are removed first, and glob semantics remain centralized in the service layer.

## Why
Previously, private files were matched and counted before response redaction. Their presence could change success versus scan-limit error behavior for a non-owner, creating a confidentiality oracle.

## Before / After
Before: a private `/memory/user` file could influence regex work and total-scan accounting even though its matches were later hidden.

After: `grep_excludes_private_subtree_before_scan_limit_accounting` proves an excluded private file neither consumes the public scan budget nor appears in results. The branch was rebased onto current `origin/main`; the focused regression test and all 10 `just pre-push` gates pass.

## Risk
- Medium
- The grep storage API changed across all backends. Rebase resolution preserves current glob semantics and adds the exclusion independently, before content access.
- Owners and trusted internal callers retain existing private-memory access.

## Security
- Threat categories reviewed: TM-TENANT, TM-FS, TM-DOS, TM-API
- Findings and resolutions: private candidates are excluded before matching, content reads, and scan accounting across every backend; response redaction remains defense in depth. SQL stays parameterized, path matching remains bounded, and no cross-session behavior changes.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated
- [x] All review comments addressed (no review comments)

### DIFF
--- a/crates/server/src/api/workspace_files.rs
+++ b/crates/server/src/api/workspace_files.rs
@@ -543,6 +543,7 @@ pub async fn grep_files(
             GrepInput {
                 pattern: req.pattern,
                 path_pattern: req.path_pattern,
+                excluded_path_prefix: None,
             },
         )
         .await
--- a/crates/server/src/direct_worker_adapters.rs
+++ b/crates/server/src/direct_worker_adapters.rs
@@ -1241,7 +1241,7 @@ impl WorkerAdapters for DirectWorkerAdapters {
                 .map(everruns_core::session_path::GrepPathPattern::new)
                 .transpose()
                 .unwrap_or(None);
-            let virtual_matches = registry.grep(&session_id, &regex, None, 512 * 1024);
+            let virtual_matches = registry.grep(&session_id, &regex, None, None, 512 * 1024);
             matches.extend(
                 virtual_matches
                     .into_iter()
--- a/crates/server/src/domains/session_files/commands.rs
+++ b/crates/server/src/domains/session_files/commands.rs
@@ -520,6 +520,8 @@ impl Command for GrepWorkspaceFiles {
                 GrepInput {
                     pattern: self.req.pattern,
                     path_pattern: self.req.path_pattern,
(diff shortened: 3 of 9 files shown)


chaliy:
Deep review — holding this one back from merge pending one gap.

The command-layer fix (threading `excluded_path_prefix` for non-owners across the SQL/in-memory/virtual-mount backends, with redaction retained as defense-in-depth) is correct and well-tested.

However, there is a **parallel grep path the fix doesn't cover**: the HTTP handler `crates/server/src/api/workspace_files.rs::grep_files` (`POST /v1/workspaces/{workspace_id}/fs/_/grep`) calls `file_service.grep(...)` directly with a plain `GrepInput` — it applies neither the new `excluded_path_prefix` exclusion nor `redact_user_memory_files` (both live only in `commands.rs`). It authorizes via `resolve_for_read` on `ResolvedOrg`, so before merging please confirm this org-scoped workspace endpoint cannot expose another participant's private-memory subtree (or thread the same exclusion/redaction through it). If it's provably owner/org-only for the private-memory trust boundary, this is a no-op and the PR is good to merge.

The rest of the aardvark batch has been rebased and is being merged as CI goes green; this PR is the only one held.

---

chaliy:
Confirmed the late review finding: the workspace-scoped HTTP grep route cannot establish session ownership. Follow-up #3025 excludes `/memory/user` before matching/reads/scan accounting and retains response redaction as defense in depth.

PR #3005, opened 2026-08-07
fix(authz): restrict Platform Chat turns to owner

chaliy:
## Summary
- require the authenticated caller to own a Platform Chat session before accepting a user message
- preserve message creation behavior for other harnesses
- cover both rejected non-owner turns and allowed owner turns

## Testing
- `cargo fmt --all -- --check`
- `cargo test -p everruns-server domains::messages::commands::tests::platform_chat --lib`


chaliy:
Superseded by #3008, which preserves trusted internal callers, validates the built-in Platform Chat harness, includes owner/non-owner regression coverage, updates the threat model, and has green CI.

PR #3010, opened 2026-08-07
fix(authz): restrict Platform Chat turns to owners

chaliy:
## What changed
Platform Chat now accepts user messages only from the persisted owner of the target session. Other org members retain the existing org-wide session management behavior for non-Platform Chat sessions.

Regression coverage verifies that an owner can start a Platform Chat turn and a different org member receives HTTP 403. The threat-model mitigation is updated at the authorization boundary.

## Why
Platform commands execute with the persisted session owner's permissions. Org-wide session management previously let another member submit a message to an owner-owned Platform Chat session and drive catalog mutations as that owner.

## Before / After
Before: source tracing showed `CreateMessage` loaded any org-scoped session and immediately started its turn, while Platform execution later resolved the persisted owner.

After: `cargo test -p everruns-server domains::messages::commands::tests::platform_chat_ --lib` passes two regression cases: the owner is accepted and a non-owner is rejected with 403.

## Risk
- Medium
- Platform Chat sessions are intentionally no longer collaborative across users. Ordinary session messaging remains unchanged.
- Message creation performs one harness lookup before routing.

## Security
- Threat categories reviewed: TM-AGENT, TM-AUTHZ, TM-TENANT, and TM-TOOL.
- Findings and resolutions: confirmed the confused-deputy path remains in HEAD. Bound Platform Chat turn creation to `resolved_owner_user_id`, kept the owner-derived worker caller intact, added negative/positive authorization tests, and updated TM-AGENT-017. No new injection, cross-tenant, secret, logging, or unbounded-work surface was introduced.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated


chaliy:
Superseded by #3008. This variant does not preserve the documented trusted-internal caller path; #3008 covers that behavior and has green CI.

PR #3054, opened 2026-08-08
fix(security): redact reflected MCP credentials

chaliy:
## What changed

- Redacts every injected MCP credential value from successful tool results before they can reach model context, events, or persistence.
- Redacts reflected credentials from JSON-RPC and HTTP transport errors before executor logging or error propagation.
- Adds regression coverage for malicious success, JSON-RPC error, and non-2xx response reflection.
- Updates TM-TOOL-029 to document response-boundary redaction.

## Why

A buggy or malicious MCP endpoint could echo a securely injected credential in tool content or an error response. The response and executor paths previously forwarded that plaintext into session-visible output, durable events, model context, or logs, violating the egress-only credential guarantee.

## Before / After

Before: source-path validation demonstrated that an MCP endpoint reflecting a bound sentinel could place it in `ToolResult` or propagated error text.

After: `cargo test -p everruns-mcp --test http_transport` passes regression cases proving the sentinel appears only in the outbound MCP request and is replaced with `[REDACTED]` in successful and failing inbound paths (11 tests passed).

## Risk

- Low
- A legitimate MCP response containing a substring exactly equal to a bound credential is redacted. This is intentional fail-safe behavior.

## Security

- Threat categories reviewed: TM-TOOL, TM-LLM
- Findings and resolutions: Confirmed TM-TOOL-029's egress-only guarantee did not cover untrusted MCP response reflection. The executor now tracks only credentials injected for the current call and scrubs all result strings and error text before any downstream observer. Empty values are ignored to avoid destructive replacement. Tests cover success content, JSON-RPC errors, non-2xx bodies, and unchanged egress injection.

## Follow-ups

No follow-ups.

## Checklist

- [x] Tests added or updated


chaliy:
Duplicate of #3062, which carries the `aardvark` label and is the branch being driven to merge.

Both branches implement the same `TM-TOOL-029` executor-boundary redaction (`redact_text` / `redact_json` / `redact_tool_result` in `crates/mcp/src/executor.rs`, plus reflection regression tests in `crates/mcp/tests/http_transport.rs`). #3062 has been rebased onto current `main` and refined, so this branch has nothing additional to contribute. Suggest closing once #3062 lands.

---

PR #3120, opened 2026-08-11
refactor(platform): move connector, OAuth, credential and system-email infrastructure out of core

chaliy:
## Summary

Removes hosted connection/authentication and concrete email infrastructure from `everruns-core`, completing the record-extraction series (EVE-877 Agent, EVE-881 Harness, EVE-882 Session, EVE-878 eval/observer/feature).

- **Connector catalog → `crates/platform/src/connector.rs`**: `Connector` trait, `ConnectorRegistry(Builder)`, `ConnectorPlugin` inventory registration, connector type/validation/form-schema values. Consumers updated across server, `crates/ard`, `crates/test-support` (fixtures feature optionally pulls platform; `sim`-only consumers unaffected), and the eight `integrations/*` crates.
- **System email → `crates/platform/src/email/`**: contract, templates, `SystemEmailConfig`, `ResendEmailSender`. The narrow `EmailSender` trait lives in the platform contract.
- **OAuth 2.1 protocol client → `crates/mcp/src/oauth/protocol.rs`**: `everruns-mcp` is its only consumer (worker MCP refresh, MCP login handshake), and `mcp → platform` would drag control-plane records into Framework/host trees — so it lands in mcp rather than platform. The server's own control-plane OAuth egress client is untouched.
- **`CredentialProvider`/`EnvCredentialProvider`/`ProviderCredentials` → `everruns-provider`**, re-exported by core at unchanged paths — provider auth contracts stay in the provider SPI with zero consumer churn. `UserConnectionResolver` (the connection-resolution seam execution genuinely needs) intentionally stays in core traits.

## Narrow services instead of the service bag

`PlatformDefinition` loses `connectors`/`email_sender`. The server composes them explicitly — `ServerAppBuilder::connector_registry(...)` / `::email_sender(...)` with OSS defaults — threaded to the connections/agent-identity API states, MCP endpoint, gRPC worker service, direct worker adapters, org invitations, `ServerContext`, and `BuiltinAuthBackend::with_email_sender`, mirroring the `with_built_in_harnesses` pattern from #3116.

## Core dependency delta

Core sheds ~2,395 lines and the `serde_urlencoded` + `eventsource-stream` direct dependencies; `cargo tree -p everruns-core` carries no platform edge on any edge kind (guard-enforced). The transitive tree shrinks only slightly (451→450 nodes) since `reqwest` remains core's egress boundary — the material win is that no OAuth/email HTTP code lives in the kernel.

## Secret redaction

Custom `Debug` + tests asserting secrets never render: `TokenSet`, `PkcePair`, `ProviderCredentials`, Resend config/sender. `TokenSet` keeps `Serialize` for encrypted persistence with an explicit doc warning against event/snapshot serialization.

## Compatibility

- **OpenAPI export byte-identical** (215 endpoints); no REST shape changes; no migration.
- `tests/fixtures/external-consumer` lockfile regenerated (dependency graph changed); evals/research/examples lockfiles verified green untouched.

## Testing

`just pre-push` green (17/17), isolation guard extended with connector/OAuth/email types. Server lib 2117 (OAuth callback/refresh, MCP connection-required, single-flight refresh, email flows), mcp/ard/platform/provider suites all green, core lib + doctests green. Rebased onto main after #3118.

Fixes EVE-879

---
_Generated by [Claude Code](https://claude.ai/code/session_01GHMMkzZ6bCUCadX8NSTLs5)_


chaliy:
CI is red here for two reasons, neither caused by this branch. Both are fixed in #3121, which is in CI now; once it lands I'll merge main in so these rerun.

- **Parallel MCP Tests** — `payments_capability_disabled_by_default`. #3118 changed core's fallback for unknown capability registration gates to the experimental rule, which auto-enables in dev, so the `machine_payments`-gated payments capability started registering by default. This branch is simply the first to touch `integrations/parallel/**`, which is what triggers that path-filtered job — the regression has been latent on main since #3118.
- **Unit Tests (Pure)** — `core_manifest_has_no_edge_to_platform`. The guard substring-matches the whole manifest, so the comment added here recording that the email senders moved to `everruns-mcp`/`everruns-platform` trips it despite there being no dependency edge. #3121 makes the guard strip comments; the comment stays.
- **Build Check** is the aggregate gate rolling those up.

---

PR #3247, opened 2026-08-21
fix(everruns): reject ambiguous MCP server names

chaliy:
## Summary
- reject MCP server names containing the reserved delimiter after sanitization
- reject server-name collisions after sanitization before routing is constructed
- cover collision and delimiter cases through the public agent builder

## Testing
- `cargo test -p everruns --test agent_builder`
- `cargo fmt --all -- --check`
- `just pre-push` (all applicable checks passed except migration immutability, because the checkout initially had no `origin/main` ref)


chaliy:
Closing because the local branch name collided with an existing remote branch; replacement follows.

PR #3264, opened 2026-08-22
fix(runtime): terminalize model-less chat turns

chaliy:
## Summary

Fixes [EVE-930](https://linear.app/everruns/issue/EVE-930/fixruntime-model-less-chat-turn-exhausts-retries-and-leaves-session).

A model-less turn was treated like a transient activity failure, retried to the DLQ, and left its workflow/session active. The Chat composer could also submit while neither an explicit nor default model resolved.

This change:

- adds a typed, non-retryable `model_not_configured` runtime error with safe user-facing localization;
- lets direct and gRPC workers distinguish retryable from deterministic task failures;
- atomically elects the worker or stale-task reaper that owns exhausted-turn terminal effects;
- records `workflow.failed`, emits `turn.failed` and `session.idled`, and returns the session to idle;
- prevents mouse, keyboard, and command submission until an enabled explicit model or resolved default exists, including stale stored selections;
- documents the failed-turn durability contract and adds the manual regression case.

## Validation

- `just pre-push` — all 22 checks passed
- `cargo test -p everruns-durable persistence::memory::tests --lib` — 36 passed
- `cargo test -p everruns-server --lib` — 2,216 passed, 2 ignored
- `cargo test -p everruns-worker --lib` — 104 passed
- `cargo test -p everruns-host runtime_context --lib` — 3 passed
- focused provider model tests — 2 passed
- Chat panel/model-selection tests — 35 passed
- UI typecheck and lint passed
- `just check-okf` — conformant

## Manual smoke

Using a fresh isolated local stack with no resolvable default model:

1. The composer showed the model-required guidance and unavailable default.
2. Enter and Send could not create a turn; the text remained in the composer and the session had no lifecycle events.
3. After enabling and explicitly selecting the simulator model, a follow-up in the same chat completed normally.
4. The session ended idle with `turn.completed` and `session.idled`.

## Security review

- No new public ingress, authorization surface, or secret handling.
- Raw durable errors remain server-side; terminal UI events use the existing safe error classifier.
- The running-to-failed compare-and-set is atomic, so worker/reaper races elect exactly one terminal-effects owner.
- Existing organization/session service boundaries are preserved.
- The protobuf addition is backward-compatible: omitted `retryable` from older workers keeps normal retry behavior.


chaliy:
### UI smoke proof

With no enabled or default chat model, the composer blocks keyboard and button submission, preserves the draft, identifies the default as unavailable, and shows actionable model guidance.

Chat composer with unavailable default model, preserved draft, model-required guidance, and disabled Send control

After enabling and explicitly selecting the simulator model, a follow-up in the same session completed and the session returned to idle.

PR #3276, opened 2026-08-25
feat(reasoning): model reasoning as ordered artifacts and publish the classification

chaliy:
## What changed

Reasoning is now a first-class, ordered part of a message instead of two flat
strings, and the classification a client needs is published instead of internal.

**For callers.** `GET /v1/sessions/{id}/messages` now carries `phase`
(`commentary` / `final_answer`) and `phase_source` (`provider` / `derived`), so a
client can finally tell an intermediate message from the answer — and can tell a
real provider classification from one the runtime inferred from tool-call
presence. Reasoning appears as ordered `reasoning` content parts with opaque
replay state stripped.

**For providers.** Each reasoning artifact keeps its own text, signature, id and
position, so replay matches what providers actually require:

- Anthropic replays every thinking block with the signature that signs *it*.
  Interleaved thinking (enabled whenever budget thinking meets tools — the normal
  agent configuration) emits several blocks per response; they were concatenated
  and paired with the last signature, which does not sign that content.
  `redacted_thinking` is handled, having been unrepresented anywhere.
- OpenAI replays every reasoning item under the `rs_…` id the provider issued
  (ids were synthesized from a counter), and requests
  `include: ["reasoning.encrypted_content"]` — without it the provider returns no
  payload, so the replay path existed but never had anything to replay.
- Gemini gains reasoning entirely: `thinkingConfig` with `includeThoughts`,
  thought-part parsing, and `thoughtSignature` capture and replay including
  per-function-call binding. The model profiles advertised `reasoning: true`
  while the driver had zero reasoning code, so the effort selector was a no-op.
- Chat Completions parses `reasoning_content` / `reasoning`, so DeepSeek-R1,
  Qwen, Groq and Fireworks reasoning is no longer read nowhere.

**Channel correctness.** OpenAI's reasoning summary was mapped to `TextDelta`,
which persisted it as the model's answer and replayed it to the model as its own
prior output. It now reaches the reasoning channel labelled as a summary.

## Why

The representation was shaped after Anthropic's 2024 wire format
(`Message.thinking` + `thinking_signature`) with every other provider squeezed
through it. Current provider requirements are about *position and identity*, not
text, and a single per-message field erases both. Above that, `phase` never
reached the API — and was in fact dropped crossing the worker gRPC boundary, so
every message arrived unclassified regardless of what the provider reported.

## Before / After

Assistant message with interleaved thinking, as replayed to Anthropic:

```
before   [ thinking("First I check the logs" + "Now I read the diff"), sig: "sig-second" ]
after    [ thinking("First I check the logs", sig-first),
           thinking("Now I read the diff",   sig-second) ]
```

GPT-5 assistant message content:

```
before   "Breaking down the problem... The build fails because ..."   (summary folded into the answer)
after    "The build fails because ..."                                (summary on the reasoning channel)
```

REST message (new fields):

```json
{ "role": "agent", "phase": "commentary", "phase_source": "derived",
  "content": [ { "type": "reasoning", "provider": "anthropic",
                 "text": { "kind": "plain", "text": "..." } } ] }
```

Gemini request now carries `generationConfig.thinkingConfig`; OpenAI requests
carry `include`. Both were absent.

## Risk

- **Medium**
- Breaking by intent (pre-1.0): `Message.thinking` / `thinking_signature` are
  gone, `LlmStreamEvent` reasoning variants are replaced, proto fields 7/8 are
  reserved, and the OpenAPI schema names `RuntimeMessage` /
  `RuntimeMessageRole` replace the previous collision.
- Reasoning replay now sends provider-issued ids and per-block signatures where
  it previously sent synthesized ones. Wire tests pin the shapes, but the live
  matrix against real providers is the real check.
- Gemini reasoning is new code on a previously untouched path.

## Security

- Threat categories reviewed: **TM-LLM** (provider artifacts, key/secret
  exposure, CoT retention), **TM-API** (published surface).
- Findings and resolutions: added **TM-LLM-034**. The surface narrows — the
  published `Message` schema carried `thinking_signature` and `reason.item`
  persisted `encrypted_content`; both are removed, `to_public()` strips replay
  state at the REST projection, ATIF skips the artifact, and driver logs record
  lengths and booleans, never payloads. One deliberate widening is recorded:
  chain-of-thought retention is now uniform across providers, so sessions on
  Chat Completions reasoning models persist reasoning they previously discarded.
  Operators who must not retain CoT should treat reasoning parts and
  `reason.thinking.*` as session content under the same retention controls as
  messages.

## Follow-ups

- **Split the `reason.*` event namespace.** `reason.started` / `reason.completed`
  mean "LLM inference step" while `reason.thinking.*` / `reason.item` mean model
  reasoning. Deferred because renaming published event types touches the UI, SDK,
  docs and every consumer, and is PR-sized on its own; the distinction is now
  documented in `docs/event-reference.md` and `knowledge/execution/events.md`.
- **Make `phase` meaningful on non-native providers.** It is currently
  `has_tool_calls`, so a text-only preamble is reported as `final_answer`.
  `phase_source: derived` makes that visible; deriving a real classification is a
  behavioral change deserving its own review.
- **Finish the projection at the events boundary (EVE-933).** `Message::into_public()`
  strips `signature` / `encrypted` on `GET /v1/sessions/{id}/messages`, but
  `GET /v1/sessions/{id}/events` returns stored rows verbatim, so
  `output.message.completed` still carries them. Not a regression — `main`
  published `thinking_signature` on that same endpoint and persisted
  `encrypted_content` on `reason.item`, both of which this PR removes — but it
  leaves one read path sanitized and one not. Deferred because projecting
  `EventData` touches the list and SSE paths plus the four other surfaces that
  parse `OutputMessageCompletedData`, and persistence must keep the state for
  replay.

- **Type `reasoning_effort` end to end.** A `ReasoningEffort` enum exists but
  `Controls.reasoning.effort` is `Option<String>` re-parsed per driver, and
  `minimal` has no budget mapping. Orthogonal to this change's shape.

## Evidence

| Check | Result |
|---|---|
| `cargo check --workspace --all-targets` | pass |
| `cargo test -p everruns-core` | 1053 passed, 0 failed |
| `cargo test -p everruns-provider` | 525 passed, 0 failed |
| `cargo test -p everruns-server --lib` | 2220 passed, 0 failed |
| engine / internal-protocol / anthropic / gemini / builtins / host | 36 suites, 0 failed |
| `./scripts/export-openapi.sh` | pass, 216 endpoints |
| `pnpm lint` / `typecheck` / `api-types:check` (UI) | pass |

`crates/durable` integration tests fail locally with
`Failed to connect to PostgreSQL: PoolTimedOut` — they need a live database this
container does not run. This change touches zero files under `crates/durable/`.

Four existing tests turned out to be pinning the old behavior in place rather
than merely referencing removed APIs, and each was inverted into a guard:

- the OpenAI summary delta asserted `TextDelta` (the mis-channelled summary)
- a workflow test read `message.thinking_signature` out of an event payload
- the `reason.item` insta snapshot recorded `encrypted_content` in the published shape
- `..._multiple_reasoning_items_get_unique_ids` asserted only that synthesized ids
  differed, which a counter satisfies while producing ids the API cannot resolve

Not smoke tested against a running stack: this container has no Doppler
credentials or provider keys, so the agent stack cannot start here. Wire-level
tests with mocked providers cover the request and stream shapes for all four
drivers, and the live matrix in `crates/llm-tests` exercises real providers in
CI — including the reasoning assertions that replaced the
`reasoning_on_thinking_field` accommodation.

## Checklist
- [x] Tests added or updated

---
_Generated by [Claude Code](https://claude.ai/code/session_01XDwmfuubczWiykh4g21Byw)_

---


chaliy:
## Smoke test status — correcting the PR body

The PR body says this was not smoke tested because the container had no Doppler
credentials. That is no longer accurate and I want to correct it rather than
leave a stale caveat.

Credentials are available. The stack now builds and runs locally with this
change: PostgreSQL, NATS, Valkey, API, worker, UI and Caddy all come up,
`/health` returns `{"status":"ok","version":"0.21.0","auth_mode":"None"}`, and
the API serves `/api/v1/...` normally.

**Verified against the running stack:**

- Session and message creation succeed.
- `controls.reasoning.effort` round-trips through the newly typed
  `ReasoningEffort`: posting `{"reasoning":{"effort":"low"}}` returns
  `"controls":{"reasoning":{"effort":"low"}}` unchanged.
- The message projection, event stream, and worker path all run without error on
  the new message shape.

**Not verified, and why:** both provider accounts are out of credits
(`credit_balance_exhausted` from OpenAI and Anthropic alike), so no LLM turn can
complete successfully here. Three assertions remain outstanding, all needing a
successful assistant completion:

- `phase` / `phase_source` populated on a successful agent message
- reasoning artifacts persisted as ordered `reasoning` content parts
- the leak checks: `signature` / `encrypted` absent from the REST projection and
  from `reason.item` events — the TM-LLM-034 claim

Those are covered by wire tests against mocked providers and by the live matrix
in `crates/llm-tests`. CI has working credentials, so that is where they will
actually run, and they are the highest-value checks still outstanding on this
PR.

One incidental finding while probing: on the provider-error path the assistant
message carries neither `phase` nor `phase_source`. That is defensible — an
error message is neither commentary nor a final answer — but it is worth a
reviewer's eye, since a client filtering on `phase` will see error messages as
unclassified.

---

chaliy:
## Review: the unclassified error-path message is correct as written

Took the error-path finding as the reviewer. Confirmed the behavior — `phase` and
`phase_source` are assigned only on the successful-completion path
(`crates/engine/src/execution/reason.rs`, where `provider_phase` resolves), so an
assistant message produced by a provider error carries neither.

Not changing it, for two reasons:

1. `None` is a documented, meaningful value here, not an omission. `events.rs`
   defines it as "not yet classified — treat as ordinary assistant text, NEVER
   thinking". An error surfaced as assistant text is exactly what a client should
   render it as, so the documented fallback already produces the right handling.
2. Error messages are already classified, by the channel that exists for it:
   `error_code` / `error_fields` / `error_disclosure` on
   `OutputMessageCompletedData`. Overloading `phase` to also mean "this is an
   error" would give one field two taxonomies, which is the ambiguity
   `phase_source` was added to remove.

Inventing a third `ExecutionPhase` variant for errors would be a published-enum
change, and it would earn its keep only if a consumer needed to distinguish an
error message from an unclassified one *without* reading `error_code`. Nothing
does today.

## Merging

CI is green on `20d0431` (54/54). Since this PR last ran, I resolved its conflict
with `main` — #3277 and #3278 landed underneath it. The only real conflict was
`knowledge/log.md`, where both sides added a `## 2026-08-25` entry; both are
kept. `docs/api/openapi.json` and the UI's generated types auto-merged textually,
which is not trustworthy for derived artifacts, so I regenerated both from the
merged source and confirmed the result is identical to the auto-merge (no path,
schema, or content difference). Clippy, `cargo fmt`, the OKF bundle check and the
migration-ordering check all pass on the merged tree.

The smoke-test caveat in the PR body is superseded by the comment above; the
three outstanding assertions run in CI, which has working provider credentials.

One follow-up filed from review: **EVE-933** — `GET /v1/sessions/{id}/events`
still publishes the `signature` / `encrypted` replay state that this PR strips
from `GET .../messages`. Pre-existing rather than introduced here (`main`
published `thinking_signature` on that endpoint and persisted `encrypted_content`
on `reason.item`, both of which this PR removes), so it is left to its own change
rather than widening this one. It is listed under Follow-ups in the body.

---

PR #3292, opened 2026-08-27
fix(test): repair the five tests the new credentials exposed on main

chaliy:
## What changed

main is red at 8a14ea2. #3288 gave the workflow-test job Doppler credentials,
and four tests that had been skipping themselves for want of an API key started
running. They fail. A fifth, the reasoning projection test added in that PR,
fails for a different reason. This repairs all five.

Nothing here changes product code. Every failure is a test that stopped
executing at some point and never noticed the API move underneath it.

## The five failures

| Test | Cause |
|---|---|
| `test_reasoning_reaches_api_sanitized_and_classified` | took the org default model; the CI server and worker start outside `doppler run` and hold no credential, so the turn came back `provider_misconfigured` |
| `test_anthropic_extended_thinking` | `POST /v1/agents/{id}/sessions` — not a route |
| `test_anthropic_extended_thinking_with_tools` | same, plus `tool.call_started` / `tool.call_completed` — not event types |
| `test_agent_execution_anthropic_with_tool_calls` | `claude-3-5-haiku-latest` retired; the turn errored, so no tool was called |
| `test_agent_filesystem_and_bash_workspace_integration` | same retired model |

Found while fixing them, all in the same two thinking tests:

- `claude-sonnet-4-20250514` produced no thinking, so the accumulator stayed
  empty and `reason.thinking.completed` never fired.
- `message.agent` is not an event type. The message events are
  `output.message.started` / `.delta` / `.completed`.
- Names were fixed strings, and the name stays claimed by the soft-deleted row
  after each test's own cleanup. These passed once against a given database and
  409ed on every rerun — including the CI re-run someone reaches for on a
  suspected flake.

## The one that is ours

The thinking tests read a flat `message.thinking` string. That field was
removed deliberately in #3276: one field per message cannot express several
thinking blocks, each with its own signature and position, which is what
providers require. Reasoning rides as ordered `reasoning` content parts, and
the tests now read those.

That is the documented pre-1.0 break, and it is the only failure the reasoning
rework caused. The rest is drift.

## Why this is worth the noise

The reasoning event stream was never broken. It looked that way at first —
`reason.thinking.completed` was missing — and the cause was a retired model
producing no thinking, not the rework. Running these tests is also what
confirmed, independently, that `output.message.completed` carries a reasoning
part while withholding replay state, which is exactly the contract #3279
established.

Four dead tests, seven stale references between them. None of it was reachable
by adding tests; only by executing the ones that already existed.

## Risk

- **Low.** Test-only. No product code is touched.
- The credentials that exposed these stay as they are: the job is push-only,
  the same trust boundary the existing live-LLM step relies on.

## Evidence

| Check | Result |
|---|---|
| `cargo test -p everruns-server --test workflow_test` (live, `EVERRUNS_REQUIRE_LIVE_TESTS=1`) | **35 passed, 0 failed** |
| the five previously failing | all pass against real OpenAI and Anthropic |
| multi-turn replay in both thinking tests | completes — thinking is sent back correctly |
| `output.message.completed` | carries a reasoning part, no replay state |

Run against a live stack with the same flag CI uses. The push build on main is
the real confirmation, since the workflow-test job cannot run on a pull request.

## Checklist
- [x] Tests added or updated

---
_Generated by [Claude Code](https://claude.ai/code/session_01XDwmfuubczWiykh4g21Byw)_

### DIFF
--- a/crates/server/tests/workflow_test.rs
+++ b/crates/server/tests/workflow_test.rs
@@ -4600,7 +4600,7 @@ async fn test_anthropic_extended_thinking() {
             API_BASE_URL, provider.id
         ))
         .json(&json!({
-            "model_id": "claude-sonnet-4-20250514",
+            "model_id": "claude-sonnet-4-6",
             "display_name": "Claude Sonnet 4 (Thinking Test)",
             "enabled": true
         }))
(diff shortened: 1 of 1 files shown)


chaliy:
Closing: redundant. #3291 landed the same repairs about nine hours before this
branch was cut, and main has been green since be52339.

I checked main once when this started and afterwards only watched #3288's CI,
so I never noticed main had recovered, and rediscovered the same four causes
independently. #3291 covers all of them — the retired session route, the retired
model ids, `message.agent`, `tool.call_started` / `tool.call_completed`, the
removed `message.thinking` field, and giving the projection test its own
provider.

#3291 is also the better version: it references the event names through the
`everruns_core::events` constants, so a future rename breaks the build instead
of silently failing an assertion. This branch hardcoded the strings.

One thing here is not in #3291, and I am not opening a PR for it unprompted:
these tests use fixed resource names, and a name stays claimed by the
soft-deleted row after each test's own cleanup, so they 409 on a second run
against the same database. CI gets a fresh database and never sees it, but a
local rerun does. If that is worth fixing, it is a small standalone change.

---

PR #3294, opened 2026-08-27
chore(models): retire sunset models and seed Bedrock Claude 5

chaliy:
## What changed

The model catalog no longer advertises models that vendors have retired, and AWS Bedrock now seeds current Claude models instead of ones that predate the driver.

Seed catalog and profile registry both drop 16 retired models:

- **OpenAI**: `gpt-4o`, `gpt-4o-mini`, `o1`, `o1-mini`, `o1-pro`, `o1-preview`, `o3-mini`
- **Anthropic**: `claude-3-opus`, `claude-3-sonnet`, `claude-3-haiku`, `claude-3-5-sonnet`, `claude-3-5-haiku`, `claude-3-7-sonnet`, `claude-opus-4-1`
- **Google**: `gemini-1.5-pro`, `gemini-1.5-flash`

Bedrock gains Claude Opus 5 and Sonnet 5, seeded disabled like every other non-default provider entry.

The Bedrock model-id format is the substantive detail here. Current Claude models have no in-region Bedrock availability at all — only the geo (`us.`/`eu.`/`au.`) and global cross-region inference profiles — so a bare `anthropic.claude-opus-5` would fail in every region. The seeds use the `global.` profiles, the one prefix offered in every commercial region and therefore the only sound default for a seed. A comment at the seed site points data-residency and GovCloud accounts at the matching geo ids instead. The driver passes the configured id verbatim to `converse_stream().model_id()`, so no prefix rewriting is involved.

`gpt-4o` was the codebase's default placeholder model, so it appeared across fixtures, doc examples, and event snapshots; those now use `gpt-5.2` (or `gpt-5.4-mini` where a small model was implied). Cost assertions were recomputed against gpt-5.2's real rates.

Deliberately left alone: `gpt-4o-transcribe` in the voice API (a distinct, current transcription model), the `id.starts_with("o1")` filter in OpenAI model discovery (it reflects what an account actually serves, not our catalog), and verbatim provider error strings in test fixtures.

### Unrelated CI fix folded in

crates.io yanked `chacha20 0.10.1` while this PR was in flight, turning **Check Dependency Licenses** red here and on `main`. Commit `e749684` bumps it to `0.10.2` in the three affected workspace-excluded lockfiles. This is not part of the model change; it is included at the author's request because it blocks every PR until it lands. See **Follow-ups** for what verifying that bump uncovered.

## Why

The catalog offered models that vendors have sunset, and the Bedrock seeds predated the current driver — anyone enabling them got entries that cannot serve traffic.

## Before / After

No runtime behavior changes for models that remain. The observable differences:

- A fresh seed no longer creates catalog rows for the 16 retired models, and creates two Bedrock Claude 5 rows (disabled) that it did not before.
- `get_model_profile()` returns `None` for the retired ids. Existing negative-path tests already assert this and were updated to say so explicitly — e.g. a retired pre-4 Claude now falls back to the client-side tool-search mechanism rather than the hosted one, which is the pre-existing unknown-model behavior.
- Existing database rows are untouched; seeding is additive and does not delete configured models.

### Smoke test

Ran the canonical local stack (`PORT_PREFIX=271 AUTH_MODE=none ./scripts/start-agent-dev.sh`) against a freshly seeded database:

- `/health` → `{"status":"ok","version":"0.22.0","auth_mode":"None"}`
- `GET /api/v1/models` returns 58 models; all 16 retired ids and both retired Bedrock ids are absent.
- Both new Bedrock rows are present and disabled: `global.anthropic.claude-opus-5` / `global.anthropic.claude-sonnet-5`. Their `profile` is `null`, the second follow-up below, observed rather than assumed.
- Surviving models keep profile-derived metadata: `gpt-5.2` → GPT-5.2, family `gpt-5.2`, 400K context; `claude-opus-5` → Claude Opus 5, 200K context.
- End-to-end turn on a surviving model (`claude-haiku-4-6-20260301`): created an agent and session, posted a message, and the agent replied — model resolution, provider dispatch, and turn execution all work.

Stack stopped afterward; no processes left running.

For the lockfile commit: reproduced the CI failure locally with `scripts/lib/check-nonworkspace-advisories.sh` (exit 1, `error[yanked]` on three crates), applied the bump, and confirmed the same script passes all four lockfiles (exit 0).

## Risk

- **Low**
- An operator with one of the retired models already configured keeps their row, but loses its profile-derived metadata (limits, cost, reasoning-effort options) and falls back to unknown-model defaults. That is the intended consequence of retiring a model, and those models can no longer serve traffic at the vendor anyway.
- Bedrock's new seeds are disabled by default, so nothing dispatches to them without an explicit opt-in.
- The lockfile bump also carries `jsonschema` 0.49.9 → 0.50.1 in two lockfiles. That is a correction, not a regression: `everruns-core` already requires 0.50.1, so those locks were unsatisfiable and `cargo check --locked` refused to build them at all. `examples/weekend-concierge-host` compiles cleanly on the new lockfile (verified locally); `evals/generic` does not, for a reason that predates this PR — see **Follow-ups**.

## Security

- Threat categories reviewed: **TM-LLM** (provider keys / model configuration), **TM-API** (public schema surface), **TM-TENANT** (seeded org-scoped data).
- Findings and resolutions: none. The diff changes catalog data, a static profile registry, doc comments, test fixtures, and three lockfiles. No credential handling, auth, SQL, tenancy scoping, or request-validation code is touched; the OpenAPI delta is example strings in field descriptions. No `THREAT[...]` markers exist in any touched file, and no new surface warrants one. No threat-model update needed. The dependency bump moves off a yanked crate and introduces no new advisory.

## Follow-ups

- **`evals/generic` does not compile, and this PR does not fix it.** `src/subject.rs:105` passes a `String` where the current API expects a `ReasoningEffort` enum. This is pre-existing drift, not caused by the lockfile bump: the error is in the crate's own source, path dependencies compile from the working tree regardless of the version recorded in the lock, and the original lockfile could not be built at all (`cargo check --locked` refused it). Workspace-excluded crates are compiled only by the weekly `examples-check.yml` sweep (Mondays 15:00 UTC), never per-PR, which is why this has gone unnoticed — exactly the EVE-663 failure mode that guard exists to catch. It will turn that sweep red on Monday unless fixed first. Left out because a source fix to an eval crate is unrelated to both the model change and the yank fix.
- **Bedrock models resolve no cost/limits profile.** The profile registry matches ids anchored at position 0 and its Anthropic descriptors are scoped to `DriverId::Anthropic`, so `global.anthropic.claude-opus-5` matches nothing and Bedrock spend is not token-costed. This is pre-existing (equally true of the Claude 3.5 Bedrock entries this PR removes) and widening the registry is a separate change with its own test surface.
- **The live LLM matrix still pins retired Bedrock ids.** `crates/llm-tests/tests/llm_test_matrix/mod.rs` uses `anthropic.claude-3-5-haiku-20241022-v1:0` and `anthropic.claude-3-5-sonnet-20241022-v2:0`. Those 3.5 ids do still resolve on Bedrock, so nothing is broken today, and repointing them at the Claude 5 profiles cannot be validated from this environment (no `AWS_BEDROCK_CREDENTIALS`) — changing untestable live-test ids risks turning the matrix red on main for no verified gain.

## Checklist
- [x] Tests added or updated
- [x] Final CI ran checks affected by this PR — all 53 checks green on `e749684`, no CI opt-out labels applied
- [x] All review comments addressed (code change or written reasoning) — no reviews or review threads were opened

### DIFF
--- a/apps/ui/src/__tests__/agent-detail-model.test.tsx
+++ b/apps/ui/src/__tests__/agent-detail-model.test.tsx
@@ -153,8 +153,8 @@ const mockModels: ModelWithProvider[] = [
   {
     id: "model-2",
     provider_id: "provider-2",
-    model_id: "claude-3-5-sonnet",
-    display_name: "Claude 3.5 Sonnet",
+    model_id: "claude-sonnet-5",
+    display_name: "Claude Sonnet 5",
     capabilities: ["chat"],
     enabled: false,
     healthy: true,
@@ -330,7 +330,7 @@ describe("AgentDetailPage - LLM Model Display in Sessions List", () => {
 
     // Model badges should not be visible
     expect(screen.queryByText("GPT-4o")).not.toBeInTheDocument();
-    expect(screen.queryByText("Claude 3.5 Sonnet")).not.toBeInTheDocument();
+    expect(screen.queryByText("Claude Sonnet 5")).not.toBeInTheDocument();
   });
 
   it("does not display model badge when model_id has no matching model", async () => {
@@ -428,7 +428,7 @@ describe("AgentDetailPage - Default Model Display in Configuration", () => {
     await renderWithSuspense({ agentId: "agent-1" });
 
     expect(screen.getByText("Default Model")).toBeInTheDocument();
-    expect(screen.getAllByText("Claude 3.5 Sonnet").length).toBeGreaterThan(0);
+    expect(screen.getAllByText("Claude Sonnet 5").length).toBeGreaterThan(0);
     expect(screen.getByTestId("provider-icon-anthropic")).toBeInTheDocument();
   });
 
--- a/apps/ui/src/__tests__/agent-detail-model.test.tsx
+++ b/apps/ui/src/__tests__/agent-detail-model.test.tsx
@@ -139,7 +139,7 @@ const mockModels: ModelWithProvider[] = [
   {
     id: "model-1",
     provider_id: "provider-1",
-    model_id: "gpt-4o",
+    model_id: "gpt-5.2",
     display_name: "GPT-4o",
     capabilities: ["chat"],
     enabled: false,
--- a/apps/ui/src/__tests__/session-detail-model.test.tsx
+++ b/apps/ui/src/__tests__/session-detail-model.test.tsx
@@ -32,7 +32,7 @@ const mockSession: Session = {
 const mockModel: ModelWithProvider = {
   id: "model-1",
   provider_id: "provider-1",
-  model_id: "gpt-4o",
+  model_id: "gpt-5.2",
   display_name: "GPT-4o",
   capabilities: ["chat"],
   enabled: false,
--- a/apps/ui/src/__tests__/session-detail-model.test.tsx
+++ b/apps/ui/src/__tests__/session-detail-model.test.tsx
@@ -69,14 +69,14 @@ describe("SessionCard - LLM Model Display", () => {
     const claudeModel: ModelWithProvider = {
       ...mockModel,
       id: "model-2",
-      display_name: "Claude 3.5 Sonnet",
+      display_name: "Claude Sonnet 5",
       provider_name: "Anthropic",
       provider_type: "anthropic",
     };
 
     render(<SessionCard session={mockSession} model={claudeModel} />);
 
-    expect(screen.getByText("Claude 3.5 Sonnet")).toBeInTheDocument();
+    expect(screen.getByText("Claude Sonnet 5")).toBeInTheDocument();
   });
 
   it("links to org-level session URL", async () => {
--- a/apps/ui/src/app/dev/file-previews/page.tsx
+++ b/apps/ui/src/app/dev/file-previews/page.tsx
@@ -226,7 +226,7 @@ const sampleJSON = {
     status: "active",
     capabilities: ["code_execution", "file_operations", "web_search"],
     config: {
-      model: "claude-3-sonnet",
+      model: "claude-sonnet-5",
       temperature: 0.7,
       maxTokens: 4096,
     },
--- a/apps/ui/src/app/dev/session-card/page.tsx
+++ b/apps/ui/src/app/dev/session-card/page.tsx
@@ -187,7 +187,7 @@ const sampleModels = {
   openai: {
     id: "model-uuid-openai",
     provider_id: "provider-openai",
-    model_id: "gpt-4o",
+    model_id: "gpt-5.2",
     display_name: "GPT-4o",
     capabilities: ["chat", "tools"],
     enabled: false,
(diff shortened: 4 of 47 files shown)


chaliy:
**Check Dependency Licenses** is failing, and the failure is not this PR's.

`chacha20 0.10.1` was yanked from crates.io. The workspace advisory check passes; what fails is `scripts/lib/check-nonworkspace-advisories.sh`, which runs `cargo deny check advisories` against the separate lockfiles of three non-workspace crates:

- `evals/generic`
- `examples/weekend-concierge-host`
- `tests/fixtures/external-consumer`

All three pin `chacha20 0.10.1` (transitively via `rand 0.10.2`). Evidence it isn't this PR's:

- This PR touches no `Cargo.toml` or `Cargo.lock` — the diff is the model catalog, the profile registry, doc comments, and test fixtures.
- The identical `chacha20 0.10.1` pin is present in all three lockfiles on `main` at its current tip (3a16d5c), unchanged by this branch. Main's last green run was 17:27Z; the yank landed after it, so main will go red on its next run too.

Not re-running the job: a yanked crate is deterministic, not a flake, so a re-run would fail identically and only burn a CI cycle.

**No fix exists yet to port** — this is currently the only open PR. The proposed patch, which I have verified locally, is a bump to the non-yanked `chacha20 0.10.2`:

```
cd evals/generic                      && cargo update -p chacha20 --precise 0.10.2
cd examples/weekend-concierge-host    && cargo update -p chacha20 --precise 0.10.2
cd tests/fixtures/external-consumer   && cargo update -p chacha20 --precise 0.10.2
```

`external-consumer` comes out as a clean one-line bump. The other two lockfiles are stale relative to the workspace, so cargo also re-syncs their path-dependency versions and updates `jsonschema` 0.49.9 → 0.50.1 (dropping `lazy_static`) as a side effect of rewriting the lockfile at all.

I am not folding this into this PR: it is an unrelated dependency fix that would widen a model-catalog change, and it belongs on `main` where it unblocks every future PR rather than just this one. Holding it for the author's call on whether it lands as its own PR.

---

chaliy:
Correction to the comment above: the author asked for the chacha20 fix to be folded in, so it is now part of this PR (commit `e749684`) rather than held back. **Check Dependency Licenses** is green as a result, and all 53 checks pass.

Verifying that bump surfaced one thing worth recording: `evals/generic` does not compile, at `src/subject.rs:105` (a `String` passed where a `ReasoningEffort` enum is expected). That is pre-existing drift, not caused by the lockfile change — the error is in the crate's own source, and the original lockfile could not be built at all. Workspace-excluded crates are only compiled by the weekly `examples-check.yml` sweep, so this will turn Monday's run red unless fixed first. Captured as the top follow-up in the PR description.

---

PR #3296, opened 2026-08-28
fix(ui): stop losing the first message of a fresh chat thread

chaliy:
## What changed

A chat thread no longer drops its opening user message. The SSE stream now takes a cursor a client with an empty snapshot can express — `after_sequence` (0 = replay from the session's first event) — and the stream handler replays from PostgreSQL for either cursor. `since_id` still covers every reconnect; the two are mutually exclusive (400), matching the list endpoint.

`698700f` additionally moves the lockfile off the yanked `chacha20 0.10.1` (transitive via `rand 0.10.2`), which is what turned `Check Dependency Licenses` red for every Rust PR — see the comment below.

## Why

A new thread has no events, so the UI's initial REST snapshot came back empty and it opened the stream with no cursor at all. The handler skips its replay phase in that case and starts live, so every event written between the snapshot and the subscription is dropped for good — for a new thread that is exactly the user's first message.

The transcript then kept the optimistic bubble for that message, and optimistic events sort to the end of the transcript by design. Net effect for the user: the first message of a chat is missing from the top of the thread and shows up under every later message, which reads as a message-ordering bug.

## Before / After

Reproduced against the SSE contract itself (`crates/server/tests/sse_replay_test.rs`): create a session, post a message, then subscribe.

Before — subscribing with no cursor after the message was written:

```
event: connected
data: {"status":"connected"}
```
(the `input.message` never arrives; still the case, deliberately, for a cursorless subscribe)

After — same session, subscribing with `?after_sequence=0`:

```
event: connected
data: {"status":"connected"}

event: input.message
id: event_...
data: {"type":"input.message", ... "text":"replay me" ...}
```

Client side, `useEvents` now asks for that replay whenever it holds no events (`use-events.test.tsx`), and still resumes from `since_id` once one is known. Full UI suite: 180 suites / 1486 tests green; `sse_replay_test` 3/3; `cargo clippy -p everruns-server --all-targets --all-features -D warnings` clean.

## Risk

- Low
- A client whose REST snapshot fails on a large session now replays up to the forward safety limit (10 000 events) instead of showing an empty transcript — more data over that one connection, and the right outcome for the user. Normal reconnects are unchanged: they carry `since_id` and never take the new path.

## Security

- Threat categories reviewed: TM-API (query surface), TM-DOS (stream cost). The new parameter is a bounded forward cursor on a session the caller is already authorized to stream; it is validated (non-negative, not combined with `since_id`) and the replay query keeps the existing 10 000-row safety limit. No auth or data-scoping change.

## Follow-ups

No follow-ups.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1176,9 +1176,9 @@ checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
 
 [[package]]
 name = "chacha20"
-version = "0.10.1"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+checksum = "65c35e4b699c7e15ccbe7ee35c005e4fc0a278d22238a2857e6ce2dadeda1b06"
 dependencies = [
  "cfg-if",
  "cpufeatures 0.3.0",
--- a/apps/ui/src/__tests__/use-events.test.tsx
+++ b/apps/ui/src/__tests__/use-events.test.tsx
@@ -25,6 +25,10 @@ jest.mock("@/lib/api/events", () => ({
   DEFAULT_EXCLUDED_EVENTS: [],
 }));
 
+const { getSseUrl } = jest.requireMock("@/lib/api/events") as {
+  getSseUrl: jest.Mock;
+};
+
 jest.mock("@/providers/org-provider", () => ({
   useOrg: () => ({ currentOrg: { public_id: "org-1" } }),
 }));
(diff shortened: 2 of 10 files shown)


chaliy:
Heads-up on the dependency line in this diff.

`Check Dependency Licenses` → *Check advisories (RustSec)* failed on `8c5c538` with `error[yanked]: detected yanked crate (try 'cargo update -p chacha20')` — `chacha20 0.10.1`, reaching us transitively via `rand 0.10.2` (`async-nats`, `bashkit`, `hickory-*`, `sqlx-postgres`, …). Nothing in this PR touches `Cargo.toml`, `Cargo.lock` or `deny.toml`: the crate was yanked upstream after that workflow last ran on `main` (2026-08-12 — it is path-filtered to Rust-dependency changes plus the weekly schedule), so any Rust-touching PR hits it now. It is not a flake either: yank state is a registry fact, so a re-run fails identically.

No other change carries the fix, so `698700f` carries it here: `cargo update -p chacha20 --precise 0.10.2` — a two-line lockfile bump to the published, unyanked patch release, same dependency set. Verified with `cargo check -p everruns-server` locally. Happy to split it into its own `chore(deps)` PR if you would rather keep this diff free of dependency changes.

---

PR #3313, opened 2026-08-31
fix(everruns): split reviewed and canonical session-event surfaces

chaliy:
## What changed

`Session::events()` now exposes two surfaces instead of one, and the second is what keeps this from being a deletion.

| Surface | Accessors | Contract |
|---|---|---|
| **Reviewed** | `SessionEventKind`, `data()`, `as_json()` | Only fields someone promoted. Stable, safe to log or forward. |
| **Canonical** | `canonical_json()` | The complete envelope, nothing withheld. Runtime-internal shape, unstable. |

Also promoted onto the typed surface, so they don't need the unstable one:

- **`ModelGeneration`** gains `model`, `provider`, `input_tokens`, `output_tokens`, `cost_usd`, `duration_ms`, `success`.
- **`narration` and `display_name`** pass through on every event carrying them — operator-authored display text.
- **`Other`** keeps identifying `event_type` but carries no payload field.

## Why

The original approach sanitized in place: `data()` and `as_json()` were narrowed and nothing replaced them. That closed the leak but destroyed data with no way to recover it — the facade exposes no other event path (`EventHistory` is not re-exported), so a consumer tracking spend or rebuilding history simply lost it. Model-generation usage went to `{}`, and `Other.payload` became permanently `null` while still existing as a field.

That wasn't hypothetical. `evals/generic` reads `SessionEventKind::Other { payload }` and grades transcripts from it.

The leak is real and worth closing. `turn.started` carries the user's prompt in `data.input_content`; tool events carry arguments and results. Passing internal payloads through by default means every field added inside the runtime becomes public API nobody reviewed, and ships wherever an application forwards the envelopes.

`knowledge/execution/events.md` already stated the rule under *Security and privacy* — *"unknown event data … is not passed through as an untyped public payload."* This is the embedding facade honoring it, with the complete record kept one explicit call away.

## Recorded by design

A new **Embedding facade surfaces** section in `knowledge/execution/events.md` records the contract, the two rules behind it (reviewed is opt-in, canonical loses nothing), and how to decide whether a field gets promoted: identity, outcome, usage, timing, and operator display text are promoted; prompts, tool arguments, tool results, and unrecognized payloads are not.

One invariant is enforced by test rather than left to convention: **the reviewed payload is a subset of the canonical one.** It may drop fields, never invent them — a synthesised field is a value no consumer could correlate with the durable record. `reviewed_data_never_exceeds_the_canonical_payload` checks every reviewed key against the canonical payload.

Every arm of `reviewed_data` names the fields it promotes; none clones `data`. That is what makes the invariant hold uniformly — see the review follow-up below.

## Before / After

Before — the prompt travels in the public envelope:

```rust
observed.as_json()["data"]["input_content"]  // "hello"
observed.data()["tool_call"]["arguments"]    // {"key": "answer"}
```

After — reviewed omits it, canonical still has it:

```rust
observed.data()                                       // {}
observed.canonical_json()["data"]["input_content"]    // "hello"
observed.canonical_json()["data"]["tool_call"]["arguments"]["key"]  // "answer"
```

Usage is reachable without touching the unstable surface at all:

```rust
SessionEventKind::ModelGeneration { input_tokens, output_tokens, cost_usd, .. }
```

Verification:

```
cargo test -p everruns --features local --lib          # 91 passed
cargo test -p everruns --features local --test session_events   # 9 passed
cargo clippy -p everruns --features local --all-targets # clean
cargo fmt --all -- --check                              # clean
cargo run --example canonical_events                    # recorded 12 canonical envelopes
cd evals/generic && cargo test                          # 22 passed
just pre-push                                           # all 22 checks, incl. Core public API guard
```

## Review follow-up

`SessionEventKind::TurnCancelled` cloned its event data wholesale, so it was the one arm where this PR's central invariant — a field added inside the runtime cannot join the public surface without being promoted — did not hold. Being a unit variant, it had no variant fields to mirror, which is what made a clone look reasonable there.

It now names `turn_id`, `reason`, `usage` explicitly. The reviewed payload is identical for every event the runtime emits today; what changes is the next field added to `TurnCancelledData`, which stays canonical-only until promoted. `a_cancellation_field_nobody_promoted_stays_off_the_reviewed_surface` pins it, and `reviewed_data`'s doc comment — which claimed it mirrors the kind field for field, never true of that arm — now states what the arms actually guarantee.

## Risk

- **Medium — breaking API change on a published crate.** `SessionEventKind::Other` loses `payload`; `ModelGeneration` becomes a struct variant; `data()`/`as_json()` narrow. `SessionEventKind` is `#[non_exhaustive]`, so `Other` losing a field and `ModelGeneration` gaining fields both require callers to update matches. That is deliberate: leaving `payload` as a permanently-`null` field would type-check, match, and silently yield nothing. Worth calling out in the `everruns` release notes.
- No data is unreachable. Every in-repo consumer that needed the full payload was migrated to `canonical_json()` and verified.
- Every promoted field is content-free by construction, and the subset invariant is tested.

## Security

- Threat categories reviewed: TM-API / information disclosure through the embedding facade.
- Findings and resolutions: the framing in the original description overstated the boundary — `Session::events()` is consumed in-process by the embedder that supplied the prompts, so returning its own data to it is not a privilege crossing. The genuine and sufficient benefit is **forward-compatibility hygiene**: without the split, each new internal field silently joins the public surface and reaches anywhere the application forwards envelopes. The change is justified on that basis rather than as a vulnerability fix.
- The canonical accessor is documented as carrying prompts and tool results and as unstable, so reaching for it is a deliberate act.

## Follow-ups

No follow-ups.

## Checklist
- [x] Tests added or updated
- [x] Backward compatibility considered — breaking, deliberate, documented above

### DIFF
--- a/crates/everruns/examples/canonical_events.rs
+++ b/crates/everruns/examples/canonical_events.rs
@@ -27,7 +27,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 Ok(None) => break,
                 Err(error) => return Err(error),
             };
-            recorded.push(event.as_json().clone());
+            // Recording for replay, so take the canonical envelope: it withholds
+            // nothing. `as_json()` carries the reviewed projection instead.
+            recorded.push(event.canonical_json().clone());
 
             match event.kind {
                 SessionEventKind::OutputStarted { .. } => print!("assistant: "),
--- a/crates/everruns/src/events.rs
+++ b/crates/everruns/src/events.rs
@@ -534,9 +534,11 @@ impl SessionEvent {
 
     /// Build the reviewed `data` payload for a projected event.
     ///
-    /// Mirrors [`SessionEventKind`] field for field. Adding a variant field
-    /// means adding it here; an event whose kind promotes nothing gets an empty
-    /// object rather than its internal payload.
+    /// Every arm names the fields it promotes — mirroring the variant's own
+    /// fields where it has them, an explicit list where it does not. Adding a
+    /// field means adding it here; an event that promotes nothing gets an empty
+    /// object rather than its internal payload. No arm clones `data`, so no
+    /// event type gains public surface without someone deciding it should.
     fn reviewed_data(kind: &SessionEventKind, data: &Value) -> Value {
         // Operator-authored display text, reviewed as safe wherever it appears:
         // a human wrote it for a timeline, so it carries no model or tool
@@ -622,9 +624,21 @@ impl SessionEvent {
                 "duration_ms": duration_ms,
                 "success": success,
             }),
-            // Reviewed safe in full: turn id, cancellation reason, and usage
-            // totals. The payload carries no conversation content.
-            SessionEventKind::TurnCancelled => data.clone(),
+            // Turn id, cancellation reason and usage totals are reviewed safe —
+            // none carries conversation content. Named explicitly rather than
+            // cloned wholesale: `TurnCancelled` promotes no variant fields, so a
+            // clone would make this the one event where a field added inside the
+            // runtime joins the public surface on its own, which is exactly what
+            // the reviewed/canonical split exists to prevent.
+            SessionEventKind::TurnCancelled => {
+                let mut cancelled = serde_json::Map::new();
+                for field in ["turn_id", "reason", "usage"] {
+                    if let Some(value) = data.get(field) {
+                        cancelled.insert(field.to_string(), value.clone());
+                    }
+                }
+                Value::Object(cancelled)
+            }
             SessionEventKind::Other { .. }
             | SessionEventKind::TurnStarted
             | SessionEventKind::TurnCompleted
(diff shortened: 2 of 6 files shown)


chaliy:
Reviewed the reworked shape. The direction is right and the earlier asks are all addressed — `canonical_json()` restores the lossless path, `Other.payload` is gone rather than left as a null hole, `evals/generic` is migrated to a real accessor, and the subset invariant is tested rather than asserted in prose.

One thing I found and fixed rather than flagged, pushed as `36804d3`:

**`TurnCancelled` was cloning its payload wholesale** (`SessionEventKind::TurnCancelled => data.clone()`). Every other arm names the fields it promotes, so that one arm was the single place where the PR's central invariant — a field added inside the runtime cannot become public surface on its own — did not hold. It is also the arm where a clone looks most reasonable, because `TurnCancelled` is a unit variant with no fields to mirror.

Now it names `turn_id`, `reason`, `usage` explicitly. The reviewed payload is identical for every event the runtime emits today; what changes is the next field added to `TurnCancelledData`, which stays canonical-only until someone promotes it. Added `a_cancellation_field_nobody_promoted_stays_off_the_reviewed_surface` to pin it, and corrected `reviewed_data`'s doc comment, which claimed it mirrors the kind field for field — never true of that arm.

On the breaking change: `Other` losing `payload` and `ModelGeneration` becoming a struct variant are both the right call over a permanently-null field, and `#[non_exhaustive]` already forces callers through a wildcard arm. Worth noting in the release notes for `everruns`, but not a reason to hold it.

Validation on the pushed head: `cargo test -p everruns --features local --lib` (91 passed) and `--test session_events` (9 passed), `cargo clippy -p everruns --features local --all-targets` clean, `cargo fmt --all -- --check` clean.

---

PR #3327, opened 2026-08-31
fix(local): serialize reconnect with wake fallback

chaliy:
### Motivation
- Close a reconnect race where a stale closed route could be pruned and the synchronous fallback turn run while a fresh host route re-registered, which could cause a turn to run underneath a live host loop instead of delivering the wake to the host.

### Description
- `WakeRoutes` keeps a `std::sync::Mutex` held only for map operations, plus a per-session in-flight *depth* for synchronous fallback turns.
- `deliver_or_claim` makes the route decision (deliver to a live host, or prune a dead one) **and** claims the fallback under that single lock, then releases it. The returned `Delivery` guard lives until the inner turn returns.
- `register` waits only while *its own* session has a claim outstanding, using a `Notify` armed before the state read so a turn finishing in the gap cannot leave it parked. No lock is held across an await, so there is no lock to re-enter and no deadlock to have — including for a turn that wakes its own session again, which increments the depth and proceeds.
- Removes the old `try_send` helper; the send/prune logic now lives in `deliver_or_claim`.

### Breaking change
`WakeRoutes::register` becomes `async` (`everruns::local`, feature-gated behind `local`). An embedder calling `routes.register(id)` must `.await` it. This is inherent to the guarantee — a reconnect that must wait for an in-flight turn has to be able to suspend. `unregister` and `live_sessions` stay synchronous. There are no in-repo callers of `WakeRoutes` outside this module.

### Testing
- `reconnect_waits_for_a_synchronous_turn_to_finish` — the original guarantee: a reconnect cannot become live while a synchronous turn for that session is running.
- `a_fallback_turn_only_holds_off_its_own_sessions_reconnect` — registration and delivery for a *second* session proceed while the first is parked mid-turn.
- `a_turn_that_wakes_its_own_session_again_does_not_deadlock` — a nested same-session wake completes rather than parking on itself.
- `cargo test -p everruns --features local --lib` — 92 passed, 0 failed.
- `cargo clippy -p everruns --features local --all-targets` clean; `cargo fmt --all -- --check` clean; migration ordering clean.

------
Originally opened from [Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a88dc8dddc0832b9c6be9afb6f3576d); reworked to the narrower per-session design after review.

### DIFF
--- a/crates/everruns/src/local/wake_routing.rs
+++ b/crates/everruns/src/local/wake_routing.rs
@@ -20,13 +20,14 @@
 //! with the host, because that part is genuinely host-specific.
 
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use everruns_core::session::ExecutionSession;
 use everruns_platform::{PlatformCreateSessionRequest, PlatformMessage};
 use everruns_provider::error::Result;
 use everruns_provider::typed_id::{AgentId, HarnessId, SessionId};
+use tokio::sync::Mutex as AsyncMutex;
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel};
 
 use super::platform_store::LocalSessionRunner;
@@ -38,7 +39,7 @@ use super::platform_store::LocalSessionRunner;
 /// or closed host does not keep swallowing wakes.
 #[derive(Debug, Clone, Default)]
 pub struct WakeRoutes {
-    routes: Arc<Mutex<HashMap<SessionId, UnboundedSender<String>>>>,
+    routes: Arc<AsyncMutex<HashMap<SessionId, UnboundedSender<String>>>>,
 }
 
 impl WakeRoutes {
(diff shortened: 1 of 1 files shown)


chaliy:
Took you up on the narrower fix. Rebased onto `main` (`bf23034`) and pushed `bec0466`.

**Global serialization — gone.** `WakeRoutes` is back to a plain `std::sync::Mutex` held only for map operations, plus a per-session in-flight depth. `deliver_or_claim` makes the route decision *and* claims the fallback under that one lock, then releases it; the returned guard lives until the turn returns. So the hold-off is per session: another session's `register`, `unregister`, `live_sessions` and `send_message` never queue behind it.

**Re-entrancy deadlock — gone, not narrowed.** Nothing is held across an await any more, so there is no lock to re-enter. A nested wake for the same session increments the depth and proceeds to its own turn. The subagent-waking-another-session shape you had in mind never touches the same entry at all. Both are pinned by new tests:

- `a_fallback_turn_only_holds_off_its_own_sessions_reconnect` — registers and delivers to a second session while the first is parked mid-turn. This one fails against the version you reviewed.
- `a_turn_that_wakes_its_own_session_again_does_not_deadlock`.

`reconnect_waits_for_a_synchronous_turn_to_finish` is unchanged and still passes, so the guarantee itself did not move.

**On the API break — you were right that it was unstated, and it is smaller now.** `unregister` and `live_sessions` are synchronous again. `register` stays `async`, and I don't think that is avoidable: the guarantee *is* "a reconnect waits", and it cannot wait without being able to suspend. Re-checking after acquiring and dropping the guard before the fallback await (your parenthetical) doesn't close it — by then the turn has already run. So the one-method break is the price of the guarantee rather than an artifact of how it was implemented. It is now stated in the commit message and the PR description, with the note that there are no in-repo callers.

Validation: `cargo test -p everruns --features local --lib` (92 passed), `cargo clippy -p everruns --features local --all-targets` clean, `cargo fmt --all --check` clean, migration ordering clean. `just` isn't on this runner, so `pre-push` ran as its individual checks.

---

PR #3543, opened 2026-09-12
fix(tests): make live workflow teardown assert, and stop it leaking

chaliy:
## What changed

Every teardown in `workflow_test.rs` now asserts that the server actually performed the delete, and the deletes happen in an order that can succeed.

`client.delete(..).send().await.expect("Failed to delete model")` only asserts the request was *sent*. A refused delete answers 500 and is still `Ok`, so teardown reported success while leaving rows in the shared test database. All 54 teardown deletes go through a `cleanup_delete` helper that checks the status (404 passes — teardown is idempotent).

That alone turned 9 tests red, which is the point: those were the silent leaks. Making them green needed two more things:

- **Sessions.** `sessions.model_id` pins the model and sessions were simply never deleted. Eight tests now delete theirs, and `test_full_agent_session_workflow` — which cleaned up nothing at all, and so could not run twice against one database — gained a teardown.
- **Migration 129.** The agent reference could not be released *at all*, which is the one thing the issue assumed but which was not true.

## Why the migration

The issue expected "delete the dependent agent before the model" to work. It cannot:

- `DELETE /v1/agents/{id}` only archives — the row survives and keeps `default_model_id`.
- `POST /v1/agents/{id}/delete` is **also** a soft delete (`status = 'deleted'`, row stays), despite being documented as "Permanently delete an archived agent". It also 404s on an already-soft-deleted agent, so it is unreachable after a `DELETE`.
- `PATCH /v1/agents/{id}` takes `default_model_id` as a plain `Option`, where null is indistinguishable from "leave unchanged".

So a model that any agent had ever selected as its default was undeletable for the life of the database, with no API able to release it. Migration 129 makes `agents_default_model_id_fkey` `ON DELETE SET NULL`, matching what `organization_settings.default_model_id` has done since 007: the model goes, the agent falls back to the org/harness default. The previous behaviour was not a useful guard — it surfaced as an opaque 500 naming neither the constraint nor the agent holding it.

`sessions.model_id` is deliberately left `NO ACTION`: it records which model a session actually ran on, and nulling that would rewrite history. Callers delete their sessions instead.

## Before / After

Verified against a live server and worker, on a fresh database each time.

**Before** (three tests, reporting green):

```
ERROR: update or delete on table "models" violates foreign key constraint
       "agents_default_model_id_fkey" on table "agents"
ERROR: update or delete on table "models" violates foreign key constraint
       "sessions_model_id_fkey" on table "sessions"

test result: ok. 3 passed; 0 failed
leaked: 2 models, 1 provider, 2 sessions, 1 agent
```

**After**:

```
test result: ok. 40 passed; 0 failed
FK errors: 0
extra_models: 0 | extra_providers: 0 | sessions: 0
```

**Run twice against the same database** — still 0 FK errors and 0 leaked models, providers or sessions.

The 15 agent rows that remain are all `status = 'deleted'` soft deletes. No API removes an agent row; that is existing product behaviour, not something teardown can reach.

Note that `Workflow Tests (Server + Worker)` — the only job that runs this file — is gated `if: github.event_name == 'push'` and so is skipped on PRs. See the comment below; the local run above is the evidence, and I will watch `main` after merge.

## Risk

- Medium
- The migration changes product behaviour, not just tests: deleting a model now clears it from agents that used it as their default instead of failing. Those agents fall back to the org/harness default. The alternative it replaces is an opaque 500 and a row nobody can delete.
- The assertions are the other half of the risk, in the useful direction: a cleanup regression now fails the test instead of accumulating rows silently.
- `just pre-push` green (22/22); migrations sequential `001..129` (renumbered from 128 after `128_files.sql` landed on main mid-review).

## Security

- Threat categories reviewed: TM-AUTHZ, TM-TENANT, TM-SQL, TM-API.
- Findings and resolutions: no cross-tenant reach. Deleting a model already requires `LLM_MODEL_MANAGE` and is org-scoped (`WHERE org_id = $1`), and an agent can only reference a model in its own org, so `SET NULL` cannot touch another tenant's rows. It does mean model-manage can now clear an agent's default model indirectly where the delete was previously refused — called out deliberately; the agent degrades to the org default rather than breaking. No new endpoint, no user input in the migration, no `THREAT[...]` marker touched.

## Follow-ups

- `test_mcp_server_crud` still cannot run twice against one database: `DELETE /v1/mcp-servers/{id}` disables rather than removes, and the name uniqueness check counts disabled rows. A different resource from the models/providers leak this issue reported, and CI runs against a fresh database. Filed as EVE-964 rather than widening this PR.

## Checklist

- [x] Tests added or updated

Fixes EVE-955

---
_Generated by [Claude Code](https://claude.ai/code/session_01HvEF7W9UNZEE8tfPLjw5Ji)_


chaliy:
Worth flagging for review: **`Workflow Tests (Server + Worker)` is skipped on this PR** — and that is the one job that runs `workflow_test.rs`, the file this PR is entirely about.

It is gated `if: github.event_name == 'push'` ([ci.yml#L1594](https://github.com/everruns/everruns/blob/main/.github/workflows/ci.yml#L1594)), because it needs Doppler provider credentials and a PR must not reach the vault. So it only runs on `main`, after merge.

The evidence for this change is therefore the local run, not CI. Against a live server and worker on a fresh database:

| | before | after |
|---|---|---|
| tests | 3 green, silently leaking | 40/40 green |
| FK errors in the Postgres log | 3 | 0 |
| leaked models / providers / sessions | 2 / 1 / 2 | 0 / 0 / 0 |

Run a second time against the same database: still 0 FK errors and 0 leaked rows. Everything else CI does run here is green.

I will watch `main` after merge and fix the job there if it comes back red.

---

PR #3569, opened 2026-09-13
fix(ui): render org settings read-only for members

chaliy:
## What changed

Members now see Settings → Organization as a read-only page instead of an editable one that cannot save.

- A notice states that only owners and admins can change these settings.
- Organization Name renders read-only; the Default Model, Default Harness and Base Harness pickers render disabled.
- No autosave badges for members — which removes both the misleading green "Saved" on an untouched page and the red "Not saved" that every edit attempt produced.
- Autosave never fires for a member, so the page stops issuing requests the backend is guaranteed to reject.
- Creating and switching organizations stay available to everyone; neither is an org-admin action.

Separately, the members page no longer fetches pending invitations for members. `GET /v1/orgs/:org/invites` is `OrgAdmin`-only, so that request was a 403 on every render (the section then silently hid itself).

## Why

`PATCH /v1/orgs/:org` is admin-only on the backend (`is_org_admin_of_public_db`), but the UI rendered every control editable regardless of role. A member could type a new org name or pick a model, watch it autosave, and get three red "Not saved" badges plus "Only organization admins can update organization settings" — with no prior signal that the page was not theirs to edit.

This is the "UI mirrors backend authorization" requirement in the access-control spec. The Features and Members pages already do this; Organization was the outlier. Org roles are unchanged — this is presentation only.

## Before / After

Smoke tested end to end on a local stack (`PORT_PREFIX=413 AUTH_MODE=none ./scripts/start-agent-dev.sh`, API `0.26.0`), flipping one membership row in Postgres between `member` and `admin` so both paths run against the same org.

**Before** — pre-fix page, viewer is `member`. Three green "Saved" badges on an untouched page; typing a name autosaved into `PATCH /v1/orgs/:org` → 403:

```
Organization Name   [Renamed By Member]
Only organization admins can update organization settings     ← red
MODELS      ✓ Saved
HARNESSES   ✓ Saved
```

**After** — this branch, same `member`, accessibility tree of the live page:

```
StaticText "Only organization owners and admins can change these settings.
            You have read-only access as a member."
textbox  "Organization Name"  : Personal Alice            (readonly, muted)
combobox "Default Model"      [disabled] : Platform default · GPT-5.6 Terra
combobox "Default Harness"    [disabled] : Generic
combobox "Base Harness"       [disabled] : Base
```

No badges rendered, and no `PATCH /v1/orgs/{org}` in the server log for the whole member session.

**Admin path unchanged** — same org, role flipped to `admin`: notice absent, controls editable, badges back, and a rename round-tripped to the database:

```
UI:  ✓ Saved
DB:  select name from organizations where public_id='org_8229…' → "Personal Alice Renamed"
```

**Invitations fetch** — server route log across the session:

```
member : (no /api/v1/orgs/{org}/invites requests)
admin  : 2 × GET /api/v1/orgs/{org}/invites   + "Invite member" button present
```

Screenshots were captured as validation evidence and are not committed to the repo.

Unit coverage: three new cases in `settings-organization.test.tsx`, verified to fail against the pre-fix page and pass after.

```
member (read-only)
  ✓ renders every org setting read-only with an explanatory notice
  ✓ never issues an update that the backend would reject with 403
  ✓ still lists organizations and allows creating one
```

Checks run: full UI suite (189 suites / 1487 tests), `tsc --noEmit`, oxlint, oxfmt, `just pre-push` (twice — before the first push and after the rebase).

## Risk

- Low
- The change is a UI gate over a backend gate that already exists, so a bug here can only over-restrict the UI, never over-permit: the server still rejects a non-admin `PATCH`. Worst case is an admin wrongly seeing read-only controls, which would require `hasRole` to be wrong — the same helper the Members and Reports pages already rely on. Verified live on both roles.
- `hasRole` returns `false` when `currentOrg` is null, so the gate fails closed. No flash of the read-only notice for admins: the page's `isLoading` already covers the org-initialization window.
- The invitations hook now returns no data for members. The only consumer renders nothing when the list is empty, which is what members already saw via the 403; admins still fetch, confirmed in the route log above.

## Security

- Threat categories reviewed: **TM-AUTHZ** (authorization) and **TM-WEB** (frontend). Checked the diff against both: no route, extractor, policy, or role check was touched, and no new trust boundary is introduced. The UI now reflects the existing `OrgAdmin` gates instead of advertising actions a member cannot perform.
- Findings and resolutions: none. Backend remains authoritative — the read-only rendering is defense-in-depth, and the spec is explicit that hiding controls without a backend check would not be sufficient on its own. No `THREAT[...]` markers are near the touched code, and no threat-model entry changes: this introduces no new threat and alters no mitigation.

## Follow-ups

Two policy questions the audit surfaced, both pre-existing and neither a UI/backend mismatch (the UI is consistent with what the backend currently permits):

- LLM Providers: `create/update/delete_provider` use `ResolvedOrg` with no admin gate, so any member can add or replace org provider credentials.
- Task webhooks (`/v1/task-webhooks/*`) are `OrgAdmin`-gated but have no UI at all.

Deferred because changing either is a permission-model decision, not a display fix, and would belong in a change that also updates the spec.

## Checklist
- [x] Tests added or updated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018vnz1zCeoRY7zCv5xCetLA


chaliy:
Smoke tested end to end against a local stack (`PORT_PREFIX=413 AUTH_MODE=none ./scripts/start-agent-dev.sh`, API `0.26.0`), with the membership row flipped in Postgres between `member` and `admin` to exercise both paths in the same org.

**Before (pre-fix page, viewer is `member`)** — reproduced the reported failure: three green "Saved" badges on an untouched page; typing a new name autosaved into `PATCH /v1/orgs/:org` → 403 and the red inline error.

```
Organization Name   [Renamed By Member]
Only organization admins can update organization settings     ← red
MODELS      ✓ Saved
HARNESSES   ✓ Saved
```

**After (this branch, same `member`)** — accessibility tree of the live page:

```
StaticText "Only organization owners and admins can change these settings.
            You have read-only access as a member."
textbox  "Organization Name"  : Personal Alice            (readonly, muted)
combobox "Default Model"      [disabled] : Platform default · GPT-5.6 Terra
combobox "Default Harness"    [disabled] : Generic
combobox "Base Harness"       [disabled] : Base
```

No badges rendered, and no `PATCH /v1/orgs/{org}` in the server log for the whole member session.

**Admin path unchanged** — same org, role flipped to `admin`, page reloaded: notice absent, controls editable, badges back. Renaming the org autosaved and persisted:

```
UI:  ✓ Saved
DB:  select name from organizations where public_id='org_8229…' → "Personal Alice Renamed"
```

**Invitations fetch** — server route log across the session:

```
member : (no /api/v1/orgs/{org}/invites requests)
admin  : 2 × GET /api/v1/orgs/{org}/invites   + "Invite member" button present
```

So members no longer fire the guaranteed 403, and admins still load pending invites.

Screenshots were captured locally as validation evidence and are not committed to the repo, per the shipping spec.

---

PR #3591, opened 2026-09-15
fix(ci): correct the stale sample-repo workspace exclude path

chaliy:
## What changed

The root `Cargo.toml` `[workspace] exclude` entry for the bashkit-repo-agent fixture repository now points at `examples/bashkit-repo-agent/src/sample-repo`, where the fixture actually lives.

This unbreaks the weekly **Examples Compile Check**, which derives its crate list from that array and treats any entry without a `Cargo.toml` as a hard error. The fixture is also now genuinely compiled by the guard for the first time since it moved.

## Why

#3552 moved the fixture under `src/` but left the exclude entry behind, so the sweep has been failing:

```
##[error]Excluded path 'examples/bashkit-repo-agent/sample-repo' has no Cargo.toml (stale exclude entry?)
One or more excluded standalone crates failed to compile: examples/bashkit-repo-agent/sample-repo(no-manifest)
```

That is the guard working as designed — it fails closed on a stale entry rather than silently checking five crates and calling it six. The sweep is weekly, so nothing on the PR hot path caught the move ([run 34861322737](https://github.com/everruns/everruns/actions/runs/34861322737), 2026-09-14).

## Before / After

Replaying the guard's own discovery step against both revisions:

| Exclude entry | Before | After |
| --- | --- | --- |
| `research/infinity_context` | OK | OK |
| `research/lua-vs-bash` | OK | OK |
| `examples/weekend-concierge-host` | OK | OK |
| `evals/platform-capability` | OK | OK |
| `evals/generic` | OK | OK |
| `examples/bashkit-repo-agent/sample-repo` | **no manifest → job fails** | — |
| `examples/bashkit-repo-agent/src/sample-repo` | — | OK |

The newly discovered crate compiles clean, which is the part that matters — the fix moves the guard forward rather than just past the error:

```
$ cargo check --manifest-path examples/bashkit-repo-agent/src/sample-repo/Cargo.toml
    Checking fetchkit-core v0.1.0
    Checking fetchkit-cli v0.1.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.07s
```

`cargo metadata --no-deps` on the outer workspace still resolves, and `just pre-push` is green (22/22).

## Risk

- Low
- One string in a manifest array. The excluded path is not a workspace member either before or after, so no member's compilation can change — confirmed by the unchanged `cargo metadata` output and by pre-push's clippy pass.
- The fixture is dependency-free (one local path dep, a 14-line lock with no external packages), so the newly enabled `cargo check` cannot fail on upstream drift and needs no committed lockfile. It resolves offline in well under a second, so the sweep's runtime is unaffected.
- No effect on the example itself: `sample_repo.rs` materializes the fixture from an explicit `include_str!` list, not a directory walk, so build artifacts left in the source tree by the guard cannot leak into what the agent is handed.

## Security

- Threat categories reviewed: no security-relevant code changes. The diff is a single path string in a build manifest — no code, dependency, network, filesystem, or authz surface is touched, and no `THREAT[...]` marker sits near a changed line. The newly compiled crate is a checked-in fixture with no external dependencies, so enabling its compile adds no new supply-chain surface.

## Follow-ups

No follow-ups.

The class of bug is already covered: deriving the list from `Cargo.toml` and failing on a missing manifest is exactly what surfaced this, so a path that moves again will fail the same way rather than be silently skipped. The only gap is latency — the sweep is weekly by deliberate cost tradeoff (documented at the top of `examples-check.yml`), and narrowing that is not worth taxing every PR.

## Checklist

- [ ] Tests added or updated — n/a; the regression surface *is* a CI guard, and the guard already detects this class of error. Verified by replaying its discovery logic against both revisions (table above) rather than by adding a test of the workflow.
- [x] Backward compatibility considered — internal build configuration only; no published or runtime surface.
- [x] All review comments addressed (none yet)

---
_Generated by [Claude Code](https://claude.ai/code/session_01XA5sFnTA8XFFrE2j3rhw8C)_


chaliy:
`Check Dependency Licenses` is red here, and it is not this PR's failure.

It is RUSTSEC-2026-0285 against rustls 0.23.44 — a newly published advisory against a dependency tree this PR does not touch (the diff is one path string in the `[workspace] exclude` array). The same check is red on `main` and on every other open PR: #3587, #3588, #3589.

Fix is up as #3592 (lockfile-only bump to rustls 0.23.45 across all four committed lockfiles). Rather than duplicate the lockfile change here, I'll rebase this branch once #3592 lands — the advisory is repo-wide, so fixing it at the source unblocks the other three PRs too.

The check this PR actually exists to fix is green. `Examples Compile Check` dispatched on this branch passed: [run 34912215081](https://github.com/everruns/everruns/actions/runs/34912215081). That is the same workflow that failed on `main` in [run 34861322737](https://github.com/everruns/everruns/actions/runs/34861322737), so the guard now discovers all six excluded crates and compiles the previously-missing fixture clean.

---

PR #3599, opened 2026-09-15
feat(environments): where a session's commands run, and what they may touch

chaliy:
## What changed

A session can now say where its commands run and what that environment can actually do, and the platform decides which orgs see it.

- **Framework.** `Environment` gains named `compute` and `containment` members, filling the extension seam its builder already documented for "process, container, or remote mount" providers. Adds the `Compute`/`ComputeSession` contract, the capability set tools and UI are assembled from, and `HostCompute` — commands on the machine the process already runs on, the target Everruns never had, behind `everruns/host-compute`.
- **API.** `GET /v1/sessions/{id}/environment` and `GET /v1/environment-targets`. A caller learns Bashkit cannot run native processes *before* the first turn instead of from a confusing tool error, and an unavailable target owes a reason.
- **UI.** An Environment panel on the session Workspace tab rendering both.
- **Feature flags.** A new `platform_managed` kind: org-scoped, so the platform enrols one tenant and not another, but never the tenant's own decision. `environments` is the first, enabled wherever sandboxes already are.

Design intent is in `knowledge/harnesses/execution-environments.md`, which also records what was rejected and why.

## Why

Choosing where a session runs meant choosing a harness. `coding-container`, `coding-daytona`, and `coding-session-sandbox` differ by one capability each and then repeat ~100 lines of near-identical prompt with provider tool names spelled into the text, while five tool namespaces (`read_file`, `bash`, `sandbox_*`, `daytona_*`, `e2b_*`) cover the same six operations. Nothing in the API or UI said what had been chosen.

Two questions were tangled: *where* code runs (Everruns' sandbox providers) and *what it may touch* (yolop's kernel policies). Neither system could ask the other's, so "run on this machine but deny the network" and "no sandbox at all" had no home.

This is the first slice: the contract, the host target, the read-only surface. Environment profiles as configuration, the machine target, and the provider ports come next.

## Before / After

**Before** — nothing in the session API mentions execution; Bashkit happens because `generic` carries `bashkit_shell`.

**After** — `GET /v1/sessions/{id}/environment`:

```json
{
  "target": { "kind": "vfs", "provider": "bashkit" },
  "containment": { "level": "isolated", "network": "deny" },
  "durability": "checkpointed",
  "capabilities": { "native_processes": false, "packages": false, "pty": false,
                    "ports": false, "portable_checkpoint": true, "network_enforced": true },
  "resolved_from": "capabilities",
  "source_capability": "bashkit_shell"
}
```

UI screenshots of the panel in all four states (Bashkit, Daytona, this machine, files-only) are attached in a comment below.

Three honesty decisions worth review:

- `resolved_from: "capabilities"` — profiles are not stored yet, so the view is derived. The field says so rather than implying config that does not exist.
- `durability` separates `checkpointed` from `provider_snapshot` by whether a recovery volume is actually configured, so nothing reports recovery it cannot deliver.
- `host` is listed as unavailable *with a reason* rather than hidden.

## Risk

- **Low.** Everything new is read-only and behind a flag that is off unless sandboxes are enabled. The one behavioral change to existing code is that `SessionSandboxService::config_for_session` now calls the shared capability resolution; its tests pass unchanged.
- What can break: the flag defaulting to on wherever `FEATURE_SESSION_SANDBOX` or `FEATURE_CONTAINER_SANDBOX` is set means sandbox deployments get the panel without asking. That is the intent (`FEATURE_ENVIRONMENTS=false` opts out), but it is the change most likely to surprise.

## Security

Threat categories reviewed: **TM-AUTHZ**, **TM-TENANT**, **TM-API**, **TM-BASH**, **TM-WEB**.

- **TM-BASH — new surface, documented.** TM-BASH-004 recorded "no real process execution" as a mitigation; `HostCompute` is the one exception, by design. Added **TM-BASH-018** stating the gating instead of leaving the older row looking absolute: it compiles only behind `everruns-host/process`, which the server does not enable, nothing in the server constructs it, no route or capability exposes it, and it reports `ContainmentLevel::None`/`Durability::None` so no profile can mistake it for contained.
- **TM-AUTHZ / TM-TENANT.** `PATCH /v1/orgs/{org}/feature-flags/platform` is `PlatformUser`-gated and limited to platform-managed flags; the tenant `PATCH` refuses those flags in both directions, so neither actor can do the other's job. `resolve_org_for_platform` deliberately skips the membership check (a cross-tenant operator is exactly who fails it) and relies on the extractor; it returns 404 for unknown orgs, consistent with TM-TENANT-002. The `PATCH` merges rather than replaces, so enrolling an org cannot clear that org's own opt-ins. Session environment reads go through the existing session-access check, so tenant isolation is unchanged.
- **TM-API.** No credentials in any response: the resolver reads only the provider name and a recovery boolean from capability config, never `provider_config` itself. Flag names are validated against the catalog before any write, so no arbitrary keys reach `org_feature_flags`. Logs carry the org public id and a user uuid, never an internal `org_id` (TM-TENANT-005).
- **TM-API, accepted.** `GET /v1/environment-targets` is not org-scoped and reveals which providers this deployment has registered. No credentials or tenant data; comparable to what the capability catalog already exposes, and behind the flag. Accepted rather than mitigated.
- **TM-WEB.** The panel renders server strings through React with no `dangerouslySetInnerHTML`.

## Follow-ups

- **Super-admin console control** for enrolling an org, in `everruns/saas`. Blocked: its `oss` submodule pins a commit predating this PR, so it cannot build against the new route until this merges. The pattern to copy is the LLM-enrolment control on the Organizations row.
- **Environment profiles as configuration** (`environment: {"use": "scratch"}` on session create), the `machine` target, kernel containment, and porting Daytona/E2B/container behind the `Compute` contract. Deferred deliberately: each needs its own slice, and profiles need a migration. All are scoped in `knowledge/harnesses/execution-environments.md`.
- **`apps/ui/AGENTS.md`** is rewritten by `next dev` on every run (a `nextjs-agent-rules` block) and has never been committed. Not touched here; a permanent fix is `agentRules: false` in `apps/ui/next.config`, which is a repo-wide call.

## Checklist
- [x] Tests added or updated

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xna6TekbERTfMH5kLR8V5T)_


chaliy:
## Validation

Rebased onto current `main` (102 commits) before validating. No migrations on this branch; `check-migration-ordering.sh` passes (001..132).

**Tests** — 5 new, all green alongside the existing suites:

| Suite | Result |
|---|---|
| `everruns-host --features process --lib compute` | 7 passed — real exec into a temp root, non-zero exit as a result not an error, timeout, honest capability pair |
| `everruns --features local --test workspace_environment` | 12 passed — containment refused in both directions, durability read off the target |
| `everruns-platform feature_flags` | 23 passed — org-scoped but platform-managed, deployment gate first |
| `everruns-server --lib environments` / `org_feature_flags` / `session_sandbox` | 6 / 4 / 6 passed |
| `everruns-server --test api_integration_test environment` | 3 passed — real router, in-memory backend |
| `apps/ui` jest | 1435 passed, 179 suites |

Also clean: `cargo fmt --check`, clippy on host/platform/server `--all-targets`, `check_okf`, UI typecheck / lint / format.

**A bug the smoke test caught.** The end-to-end tests were written because the canonical stack could not start here (no Doppler, no `just`, no Postgres in this container). They immediately failed with `feature_not_enabled` — `Ctx::minimal` leaves `feature_flags` at their default, which reads as *every feature disabled*, so both environment routes would have returned 404 in **every** deployment, not only in tests. Every other flag-gated route already passes `org.feature_flags`; this one did not. Fixed in `9c19088`, and the tests now pin it.

That is also the honest note on smoke testing: the documented startup contract (`./scripts/start-agent-dev.sh`) needs Doppler and `just`, neither of which exists in this container. Rather than skip the requirement, the flows are covered by integration tests that boot the real router and assert what a caller actually receives, plus the UI rendered in a browser below.

**UI** — the panel was rendered in Chromium against a real `next dev` server at `/dev/session-components` (a showcase section added in this PR) and reviewed in all four states:

| State | Renders |
|---|---|
| Bashkit | `Virtual filesystem` · `bashkit` · `Isolated` · `network deny`; **Native processes — no**; Recovery: Checkpointed |
| Daytona + recovery volume | `Managed sandbox` · `daytona`; all six capabilities yes; Recovery: Checkpointed |
| This machine | `Uncontained` in the destructive variant; everything runnable; **Portable checkpoint — no**; Recovery: **Not recoverable** |
| Files only | `No compute` alone, no capability table, and deliberately **no containment badge** |

The rendered screenshot could not be attached: `upload-screenshot.sh` is rejected by this environment's egress proxy (`422 — Request bodies must declare Content-Type: application/json`, which a binary upload cannot satisfy). Flagging that rather than quietly dropping the media the template asks for — happy for a reviewer to pull the branch and hit `/dev/session-components`, which is exactly what that showcase is for.

Two UI bugs the showcase exposed, both fixed in `6c9c5af` with tests: a failed background refetch hid a panel that already had data, and a files-only session showed a red "Uncontained" badge for a session that runs nothing at all.


---

PR #3602, opened 2026-09-15
fix(server): let a Slack channel exist before its credentials do

chaliy:
## Why
[EVE-1015](https://linear.app/everruns/issue/EVE-1015) made Slack first-run setup circular because a manifest required a channel and channel creation required credentials that Slack provides later.
## What
- Allow Slack channels to exist before the signing secret and bot token are configured.
- Reject every inbound Slack challenge and event until a non-empty signing secret is configured.
- Keep credential flags accurate and preserve manifest event subscriptions through the first-run flow.
## Verification
- Focused `everruns-platform` Slack credential-default and malformed-config tests; focused server config and signature tests.
- `cargo test -p everruns-server --test slack_integration_test test_slack_first_run_order_without_placeholder_credentials -- --exact --nocapture --test-threads=1`
- `cargo fmt --all -- --check`, targeted Clippy with `-D warnings`, and `python3 scripts/check_okf.py knowledge`


chaliy:
**Build Check failed on [run 11779](https://github.com/everruns/everruns/actions/runs/34932746968) — not this PR's diff, and not re-runnable.**

That run was triggered by `warp-factories[bot]` adding the `factory:foreman` label at 05:26:31, not by a push. `factory:foreman` is not in the `ci:skip-*` allowlist in `.github/workflows/ci.yml`, so the event gate resolved `run_ci=false`, all 26 jobs skipped, and Build Check exited 1 by design rather than report a green check behind a run that executed nothing (the EVE-939 guard). The step's own log says it: `run_ci=false: this run executed no jobs, so it has nothing to attest`, and `the real run's own Build Check is the one that carries the verdict`.

I have not re-run it. The gate is derived from the triggering event, which is permanently `labeled`, so a re-run fails identically — it would burn a CI cycle to reproduce a designed outcome. Any future label change on this PR will produce the same red check.

The run that carries the verdict is [run 11776](https://github.com/everruns/everruns/actions/runs/34932289799) (push-triggered, same head `c522b88`), still in progress. License Check and CodeQL are already green on this commit. I'm watching it and will fix anything it turns up.

Separately: this PR was converted to **draft** by the same bot, which blocks merge. I did not convert it and I am not flipping it back — that is the author's call.

---

chaliy:
⚠️ **`9fdc6ea` re-breaks EVE-1015. It relocates the circular dependency rather than removing it.**

That commit deletes the `url_verification` bootstrap, so `verify_slack_signature` now runs unconditionally and returns `Err` for an empty secret. Every request to a credential-less channel gets 401 — **including Slack's handshake**.

Walk the flow that leaves:

1. Create app + credential-less Slack channel → works
2. Publish → works
3. Fetch manifest → works
4. Click **Create Slack App** → Slack opens create-from-manifest
5. Slack saves the manifest and calls `request_url` with a `url_verification` challenge → **401**
6. Slack refuses to create the app → the operator never receives a signing secret → they can never reach step 8

The operator still cannot complete first-run setup without inventing placeholder credentials. The dependency moved from channel-creation to app-creation; it did not go away.

This contradicts three statements this repo already makes, none of which `9fdc6ea` changes:

- `handle_slack_manifest` doc comment: *"Slack verifies `request_url` when the manifest is saved, which is why this endpoint serves only published apps: the webhook has to be answering before the Slack app is created from the manifest."*
- `slack-setup-guidance.tsx` (reordered by EVE-970): *"Slack checks this URL the moment you create the app, so it has to be live first."*
- `knowledge/test-cases/ui/slack_app/TC001_slack_app_setup.md`, **untouched by this commit** — Step 6: *"Slack sends a `url_verification` challenge to the request URL as part of saving."* Step 7 expects `webhook_verified_at` recorded. Credentials are not pasted until **Step 8**.

EVE-1015 explicitly names TC001 as the acceptance bar: *"its Step 2/5 assertions are what should pass once this is fixed."* Under `9fdc6ea`, TC001 Step 6 cannot pass.

The commit also inverts the regression test to match the new behavior — `test_slack_first_run_order_without_placeholder_credentials` now asserts the handshake returns 401 and that `webhook_verified_at` is *not* recorded, then performs the successful handshake *after* credentials are configured. That ordering cannot occur in reality, because the credentials only exist once Slack has already created the app, which requires the handshake to have succeeded. The test passes; the user journey does not.

**The underlying concern is legitimate.** An unauthenticated endpoint that echoes a challenge and sets `webhook_verified_at` is real exposure, and I documented it as a deliberate, narrow tradeoff rather than an oversight. But "fail closed" here closes the door the feature exists to open.

If the unauthenticated handshake is unacceptable, the fix is to authenticate it rather than remove it — e.g. mint a per-channel single-use bootstrap token, carry it in the manifest's `request_url` query string (Slack preserves it and echoes it back on the challenge), and accept an unsigned `url_verification` only when that token matches and `webhook_verified_at` is unset. That keeps the endpoint closed to arbitrary callers *and* lets Slack complete manifest verification.

I have not reverted anything. Flagging for the author to decide between reverting `9fdc6ea`, or taking the bootstrap-token approach.

---

PR #3621, opened 2026-09-16
feat(apps): move publish to the endpoint, add an agent exposure suspend

chaliy:
## What changed

Ingress now reads `agent_endpoints.status` alone. **Publishing one endpoint moves exactly one endpoint.**

Liveness is resolved in a single helper, `app_ingress::endpoint_liveness`:

```
live(endpoint) = endpoint.status == live
              && agent.status == active
              && !agent.exposures_suspended
```

All seven ingress gates — Slack, AG-UI, A2A invoke, A2A agent card, FCP, webhook, public chat, api_endpoint — go through it. The agent-level terms are folded in **at resolution time** and never written onto endpoint rows, so archiving an agent or flipping the new `exposures_suspended` switch leaves per-endpoint state intact and restoring is exact.

New API: `POST /v1/apps/{app_id}/channels/{channel_id}/publish|unpublish` and `POST /v1/agents/{agent_id}/exposures/suspend|resume`. Agent exposure state (`exposed`) is derived from the endpoint rows on every read, never stored.

The App publish switch keeps working while the App domain exists: it now drives the endpoints it owns — raising only the ones left enabled, lowering only the live ones — so a disabled endpoint survives a publish cycle.

## Why

EVE-1007. The old gate was the two-dimensional `App.status × AppChannel.enabled` matrix, which can express "published App, disabled channel" and "draft App, enabled channel" — neither of which means anything to a user.

More importantly it was **too coarse to be safe**. EVE-970 sequenced publish *before* Slack app creation, because the manifest has to be fetchable before Slack will create the app from it. So obtaining a Slack manifest required publishing the whole App, which simultaneously turned on every other channel attached to it — including an anonymous public chat endpoint sitting next to it. Publishing to get a manifest should not put a public chat surface on the internet.

## Before / After

The sibling-isolation test is the point of the change. Two endpoints on one agent, `api_endpoint` and `a2a`:

**Before** — reaching either meant publishing the App, which exposed both.

**After** — publishing one leaves the other exactly as it was:

```
publishing_one_endpoint_does_not_expose_its_sibling ... ok
  publish(api_endpoint) → api_endpoint = live, a2a = draft
                        → GET /v1/e/{a2a}/a2a/.well-known/agent-card.json = 404
  publish(a2a)          → a2a agent card = 200, api_endpoint still live
```

Full suite, all green:

```
endpoint_publish_test                      6 passed
agent_endpoints_migration_test            10 passed
app_api_integration_test                   7 passed
app_a2a_integration_test                  32 passed
app_invocation_channels_integration_test   6 passed (2 pre-existing ignored)
api_integration_test                     121 passed
repository_conformance_test                6 passed
everruns-server --lib                   2418 passed
```

`just pre-push` passes all 22 checks; workspace `cargo clippy --all-targets --all-features -D warnings` is clean.

## Risk

- **Medium.** This changes what every ingress gate reads. Migration 136 re-asserts `status` from the `App.status × enabled` pair it replaces and **aborts if any row disagrees**, so no endpoint silently changes reachability at deploy time.
- Behavior worth naming: the `App::*_channel()` accessors deliberately keep filtering on `enabled`, not liveness. Slack **delivery recovery for sessions that are already running** resolves a channel through them, and gating that on publish state would orphan in-flight deliveries. Liveness lives only in the ingress gate. This was caught by `recover_uses_session_app_id_instead_of_slack_app_tag` after an earlier draft folded liveness into the accessors.
- Re-enabling a channel on a published App returns it to `live`, matching the pre-change effective gate; the in-memory backend derives this from the owning App exactly as PostgreSQL does.
- Two tests in `repository_integration_test` fail locally on PostgreSQL 16 (`agent_synthesis_migration_*`). Pre-existing, unrelated, reproduces on `origin/main`, tracked as EVE-1018. CI runs PostgreSQL 17 where they pass.

## Security

- Threat categories reviewed: **TM-AUTHZ-006**, **TM-TENANT-002**, **TM-AUTHZ-005**, **TM-TENANT-012**.
- **Gate ordering is preserved**: liveness resolves *before* auth on every path, so a caller cannot distinguish a misconfigured endpoint from credential probing. `non_live_rejections_are_indistinguishable_from_not_found` asserts that a draft endpoint, a suspended agent, and a channel id that was never issued all return the identical response.
- The new `NotLive` reasons are logged server-side only and never surfaced; they exist so operators can tell the cases apart in logs while callers cannot.
- The A2A agent card is now withheld for a non-live endpoint — it advertises the invocation URL and security scheme, so serving it for a draft or suspended endpoint leaked a surface that refuses traffic.
- `PublishEndpoint`/`UnpublishEndpoint` take a bare channel id, so they do the org-scoped parent fetch and `channel_row.app_id == app.id` assertion that TM-TENANT-012 requires, and carry the invariant comment.
- The threat-model entry for TM-AUTHZ-006 was rewritten to describe the new gate.

## Follow-ups

- **UI affordance for the suspend switch is deferred to EVE-1009**, which owns the Integrations tab and whose design already places it: "Header: agent identity control, and the suspend-all-exposures switch" (`knowledge/integrations/agent-exposure.md`). The API and derived `exposed` state this PR adds are what that tab renders.
- `App.status` stays in the schema; the deletion phase (EVE-1011) removes it. Nothing reads it for ingress any more.
- The Slack manifest route's own re-homing stays with EVE-1008, as the issue scopes it.
- Un-archiving an agent is not an available transition today, so the "restores the previous live set" property is asserted by restoring the agent directly and observing that archiving rewrote no endpoint rows — which is the invariant that makes restoration exact.

## Checklist

- [x] Tests added or updated

Fixes EVE-1007

https://claude.ai/code/session_01CkLFynXKRYoakWXyBRqe7f

---
_Generated by [Claude Code](https://claude.ai/code/session_01CkLFynXKRYoakWXyBRqe7f)_


chaliy:
Standing down on one CI failure — it is not this PR's, and there is no fix to port yet.

**Failing check:** `Integration Tests (PostgreSQL Shards) (domain)` → `plugins_remote_test`, two tests:

- `test_url_marketplace_sync`
- `test_github_marketplace_end_to_end_install`

```
called `Result::unwrap()` on an `Err` value: CommandError { kind: BadRequest(
  "URL validation failed for 'https://example.com/plugins/marketplace.json':
   Blocked host: example.com (private/internal address)") }
```

**Why it is not this PR's.** These tests serve their catalogs from an in-process fake HTTP client, so they never make a real request — but the marketplace path still runs the SSRF guard first, and `validate_url` (`crates/provider/src/url_validation.rs`) does a **live DNS lookup** of the host before deciding. The test's `example.com` therefore depends on what the runner's resolver returns; on this run it resolved to an address the guard classifies as private/internal, and the request was rejected before the fake was ever consulted.

Nothing in this diff touches plugins, marketplaces, or URL validation. The same two tests pass on `main` — [run 35093074933](https://github.com/everruns/everruns/actions/runs/35093074933) on `9a46fc3`, twenty minutes before this one — and pass locally on this branch (`plugins_remote_test`: 8 passed, 0 failed).

**Proposed patch, not applied here** because it would widen this PR into the plugins domain: `validate_url_with_resolver` already takes an injectable resolver "for unit testing", but the marketplace path calls `validate_url`, which hard-wires `default_dns_resolve`. Threading the test's fake through to the resolver — or having the fake HTTP client carry a resolver stub — makes these tests independent of runner DNS. Using a different literal host only moves the dependency. Tracked as **EVE-1021**.

Re-running the failed job once, which is the one re-run this warrants. Everything else on this head is green, and I'll keep the PR watched until it is.

---

PR #3637, opened 2026-09-17
feat(ui): ship dark mode with a user-selectable theme

chaliy:
## What changed

Everruns UI now has dark mode. A **Theme** submenu in the sidebar user menu offers Light / Dark / System; the choice persists across reloads and sessions, and "System" tracks the OS preference live.

- Dark is class-based (`.dark` on `<html>`), driven by an `everruns_theme` cookie so the server render emits the right class and navigation never flashes white. A pre-paint script in `<head>` handles the "system" case the server cannot know.
- `color-scheme` is declared on both token blocks, so scrollbars and native controls follow the theme.
- Hardcoded neutrals that would not have followed the switch (`text-gray-500`, `bg-gray-300`, `bg-zinc-500`) are now semantic tokens. Scrim overlays (`bg-black/50`), text on saturated fills (`text-white` on `bg-destructive`), and the white canvas behind user file previews stay as they are on purpose — all three are documented in `apps/ui/DESIGN.md`.

No new dependencies; the provider is ~80 lines against the existing cookie pattern.

## Why

The dark token set has been sitting in `design-system.css` since the design system landed, with nothing to turn it on.

Worse, it was half-on. Tailwind v4 defaults the `dark:` variant to `prefers-color-scheme`, and the repo never declared `@custom-variant dark`. So the 63 `dark:` utilities already written into the shadcn primitives and app components *were firing* for anyone on a dark-OS machine, while the base token block stayed light — dark-tinted button and badge overlays on a light app. Making the variant class-based fixes that as a side effect.

## Before / After

Screenshots captured against the local stack (`PORT_PREFIX=271 AUTH_MODE=none`) — see the PR comment below.

Automated smoke test against that running stack, driving the real user menu:

```
PASS  light OS -> no .dark — rgb(250, 250, 250)
PASS  menu -> dark applied — rgb(17, 19, 23)
PASS  dark survives reload — rgb(17, 19, 23)
PASS  cookie persisted — "dark"
PASS  dark OS + no cookie -> dark — rgb(17, 19, 23)

5/5 checks passed
```

Computed tokens under `.dark` resolve as the design system documents (`--background: 220 15% 8%`, `--primary: 0 0% 95%` inverting to near-white so actions stay prominent).

## Risk

- **Low**
- The blast radius is every page, but the mechanism is one class on `<html>` and light mode is byte-identical to before (verified by screenshot and by the fact that no light-mode token changed).
- The realistic failure is a *cosmetic* one: a surface nobody looked at reads poorly in dark. The 353 raw status colors (`text-green-600`, `text-amber-500`) still need a pass — see Follow-ups.
- The pre-paint script is the one ordering-sensitive piece; if it were ever blocked, the symptom is a flash of light for system-mode users, not a broken app.

## Security

- **Threat categories reviewed:** TM-WEB (frontend/XSS/CSP/cookies). No auth, authz, tenancy, SQL, filesystem, tool, or LLM surface is touched — the diff is UI styling plus one cookie.
  1. **Inline `<script>` vs. CSP — investigated, no violation.** `BASE_CONTENT_SECURITY_POLICY` (`crates/server/src/app_builder.rs:286`) is `script-src 'self'` with no `'unsafe-inline'`, which would block the pre-paint script. It does not apply here: Caddy routes only `/api/*`, `/oauth/*`, `/mcp*`, `/.well-known/*` and friends to the Rust server, while the UI document is served by Next.js and carries no CSP (`infra/railway/caddy/Caddyfile`). Next.js emits its own inline hydration scripts on that document, so any future `script-src` there must introduce nonces regardless — recorded as a comment in `src/lib/theme.ts` so whoever does that knows to nonce this script too. Deliberately did **not** loosen the server CSP: TM-WEB-010 depends on `script-src 'self'` holding for `about:srcdoc` file previews.
  2. **Script content.** `THEME_SCRIPT` is a module constant with no request data interpolated into it. It reads the cookie only to compare against `"dark"`/`"light"`; the value is never evaluated, and never written to the DOM.
  3. **Cookie hardening.** The value is concatenated into a `document.cookie` string, so `persistThemeMode` re-parses through `parseThemeMode` rather than trusting its argument — an unvalidated caller cannot append cookie attributes (`; domain=…`). Covered by a regression test. The cookie is intentionally not HTTP-only (the pre-paint script must read it) and holds no secret — a theme name. `SameSite=Lax` matches the other app cookies; TM-WEB-003 concerns the token cookies and does not apply.
- No `THREAT` markers sit near the changed code, and no new threat surface was introduced, so `knowledge/security/threat-model.md` is unchanged.

## Follow-ups

- **Status-color pass for dark mode.** 353 raw palette utilities (`text-green-600`, `text-red-600`, `text-amber-500`) across 72 files. Many already carry `dark:` pairs, but not all, and the 600/700 shades are muddy on charcoal. Deferred because the right fix is semantic `success`/`warning`/`info` tokens in the design system rather than sprinkling `dark:` variants file by file — a design decision that deserves its own PR, not a rider on this one. Filing as an EVE issue.
- **Visual sweep of dense surfaces** (chat tool cards, code blocks, file browser, evals matrix) once those tokens exist. The surfaces I could reach are verified; I did not seed data for every one.
- **Persist the theme per account** rather than per browser. Deliberately out of scope — that touches the user-settings API, and the cookie covers the common case.

Everything else in scope is in this PR.

## Checklist
- [x] Tests added or updated — 7 new tests in `apps/ui/src/__tests__/theme-provider.test.tsx` (explicit modes, OS fallback, live OS change, cookie persistence, cookie-injection guard); full UI suite green at 1562 tests
- [x] Backward compatibility considered — no cookie means "system", which reproduces today's behavior for anyone on a light OS; internal UI code, no external contract

https://claude.ai/code/session_01XEHCC9qBNLrb5JFAUhrkC3

---
_Generated by [Claude Code](https://claude.ai/code/session_01XEHCC9qBNLrb5JFAUhrkC3)_


chaliy:
## Validation evidence

Screenshots were captured but **could not be uploaded** from this session — GitHub's user-attachments CDN is unreachable through the sandbox's egress proxy (HTTP 422 on the multipart upload; `agent-browser` is also not installed here, so I drove Chromium via Playwright directly). The PNGs went to the author to attach manually. Everything below is measured rather than eyeballed, so the gap is cosmetic evidence only.

### Smoke test — real stack, real user menu

`PORT_PREFIX=271 AUTH_MODE=none ./scripts/start-agent-dev.sh`, then a Playwright script that clicks through the sidebar user menu → Theme → Dark, exactly as a user would:

```
PASS  light OS -> no .dark — rgb(250, 250, 250)
PASS  menu -> dark applied — rgb(17, 19, 23)
PASS  dark survives reload — rgb(17, 19, 23)
PASS  cookie persisted — "dark"
PASS  dark OS + no cookie -> dark — rgb(17, 19, 23)

5/5 checks passed
```

The reload case is the one worth calling out: it proves the cookie round-trips through the server render and the pre-paint script, which is what keeps dark-mode users from getting a white flash on every full page load.

### Computed tokens under `.dark`

```
html.class = dark
--background: 220 15% 8%
--foreground: 0 0% 95%
--primary: 0 0% 95%
--primary-foreground: 220 15% 8%
--card: 220 15% 10%
--muted-foreground: 0 0% 65%
--border: 220 15% 18%
```

These match `DESIGN.md` exactly, including `primary` inverting to near-white. I sampled every badge and button on `/durable` to confirm nothing renders low-contrast — the light-filled `HEALTHY` badge is `bg-primary` behaving as documented, not an inversion bug.

### Pages checked in both themes

`/chats`, `/durable`, `/durable/workers`, `/models` — the routes carrying the neutrals this PR retokenized. Light mode is unchanged (no light-mode token was touched).

### Static checks

`format:check`, `lint` (no new warnings), `typecheck`, `pnpm build`, `design:lint` (0 errors), and the full Jest suite at **1562 passing**. Re-run after rebasing onto latest `main`; `check-migration-ordering.sh` reports `001..138` sequential, and this PR adds no migration.

---

PR #3651, opened 2026-09-17
fix(mcp): complete service-owned MCP grant authorization

chaliy:
## What changed
- Service OAuth starts identity creation only after discovery succeeds, then binds the callback to the exact active agent and identity with an atomic conditional write.
- Distinct users share the encrypted identity grant; refresh rotates it, `invalid_grant` revokes it, and agent deletion removes it.
- Service MCP events preserve the human initiator, derive the identity actor, and fail closed when provenance cannot be resolved.

## Why
`actsAs: service` could read identity connections, but it had no complete authorization, refresh, revocation, or provenance lifecycle. Refs EVE-1030.

## Before / After
Before, service attachments remained `connection_required`; after authorization, the PostgreSQL lifecycle proves authorize, MCP use, refresh, cross-process decryption, and revoke for distinct users. All 59 focused tests, all-target server Clippy, and all 25 `just pre-push` checks pass.

## Risk
- **Medium.** Identity-owned OAuth authorization, refresh, revocation, and attribution change. User and session grant modes stay unchanged.

## Security
- Reviewed TM-AUTHZ, TM-TENANT, TM-TOOL, TM-API, TM-CRYPTO, TM-SQL, TM-OBS, and TM-DOS. Exact-agent callback races and provenance lookup failures now fail closed.

## Follow-ups
No follow-ups.

## Checklist
- [x] Tests added or updated


chaliy:
CI is red on `cba4954`. `Build Check` failed, but it is an aggregate — the underlying cause is **`Unit Tests (Pure) (facade-cli)` being cancelled at the 30-minute job timeout**, not a test failing. Its `Run pure unit tests (facade-cli)` step never completed.

Evidence that this is not this PR's change, gathered before re-running:

| Run | facade-cli | host-mcp | provider-engine | llm-core |
| --- | --- | --- | --- | --- |
| main `4323f86` | **16:55 ✅** | 4:23 | 3:50 | 10:59 |
| this PR `cba4954` | **35:00 ⏹ cancelled** | 5:16 | 4:32 | 12:20 |

The other three shards are 15–20% slower than main, which is ordinary runner variance. facade-cli going from under 17 minutes to past a 30-minute timeout is not the same shape.

The diff is confined to `crates/server` (7 files: the authorize/callback identity mode, the lifted `agent_identities::lifecycle` module, resolver tests, and a knowledge test case). This shard runs `cargo test -p everruns --all-features`, a no-default-features check, the `everruns-host` examples, and three `everruns-cli` integration tests — none of which execute the changed code, and all of the new behaviour is behind an opt-in `mode=identity` that nothing in those tests requests. The new tests I added live in `everruns-server`'s lib suite, which this shard does not run.

Locally on this branch: full `everruns-server` lib suite 2450 passed / 0 failed / 0 warnings, and `just pre-push-full` passes all 22 checks with none skipped.

I have re-run the failed jobs once to confirm. If facade-cli times out again, the second failure is real and I will root-cause it rather than re-run again.

chaliy:
Correcting my previous comment: **the re-run I triggered did not actually retest anything**, so the second red is not a second genuine failure.

I used `rerun_failed_jobs`, which re-runs jobs whose conclusion is `failure`. The root cause here was a job whose conclusion is `cancelled` (the facade-cli shard hitting its 30-minute timeout), and `cancelled` is not `failure` — so that shard was never re-executed. Every job in run `35291451198` shows `skipped` except `Detect Changes`, the integration aggregate, and `Build Check`, which simply re-read the same stale `needs` result and failed again for the same reason. My mistake in picking the tool, not new information about the code.

One more piece of evidence against this being the diff's fault, which I did not have earlier — the same shard on my other `crates/server` PR:

| Run | facade-cli |
| --- | --- |
| **#3649** (EVE-1029 — touched `crates/server`, `crates/core`, **and** `crates/worker`) | **13:33 ✅** |
| main `4323f86` | 16:55 ✅ |
| **#3651** (this PR — `crates/server` only, smaller diff) | **35:00 ⏹ cancelled** |

That rules out the one mechanism by which this PR could plausibly have slowed the shard. If touching `crates/server` invalidated enough cache to push a marginal job past its timeout, #3649 — a strictly larger diff across three crates — would have been slower, not two and a half times faster. It was green well inside the limit.

I have re-run the full workflow so the cancelled shard actually executes. That is the first re-run that tests the hypothesis at all. If facade-cli times out again on a real execution, I will treat it as this PR's and root-cause it rather than re-running a third time.

chaliy:
My previous comment was wrong, and this one supersedes it.

**The first re-run did work.** Run [35202089420](https://github.com/everruns/everruns/actions/runs/35202089420) attempt 2 on `cba4954` completed **green** — facade-cli passed. So the 30-minute timeout was a genuine flake, which is what my first comment predicted, and `rerun_failed_jobs` did re-execute the cancelled shard. I claimed otherwise because I inspected the wrong run.

**What I actually looked at was a different run.** [35291451198](https://github.com/everruns/everruns/actions/runs/35291451198) is a *label-event* run, created when this PR was labelled `factory:foreman`. `.github/workflows/ci.yml` resolves `run_ci=false` for label events other than the `ci:skip-*` ones, so every job skips and `Build Check` fails with `run_ci=false: this run executed no jobs, so it has nothing to attest`. That is the gate behaving as designed, not a code failure — and re-running it, as I then did, could never have gone green. Both of those reds were noise I misread as signal.

Net: there was one real failure (a flake), one successful re-run that fixed it, and two red aggregates that were artifacts.

**Current state has moved on.** `warp-factories[bot]` has merged latest `main` into this branch as `3a72168`, and CI [35291663355](https://github.com/everruns/everruns/actions/runs/35291663355) is running against that head. The green run above was for `cba4954`, so it no longer describes the PR tip; the new run is the one that counts.

Since another agent is now pushing to this branch, I am not going to force my version over that merge. If the factory is driving this PR to completion, it should keep it. If not, say so and I will pick it back up — the change itself is unmodified beneath the merge, and the local evidence for it still stands: full `everruns-server` lib suite 2450 passed / 0 failed, and `just pre-push-full` clean at 22/22 with none skipped.

chaliy:
`Provider Isolation Guard` failed on `3a72168`. **The tree is clean — I could not reproduce it, and I think the guard swallowed an infrastructure error rather than finding a violation.**

Verified locally:

| Tree | `scripts/lib/check-provider-isolation.sh` |
| --- | --- |
| `origin/main` | passed |
| `3a72168` (this PR head) | passed |
| `refs/pull/3651/merge` → `c016316`, the exact tree CI checks out | **passed** |

Likely mechanism, from the job shape and the script:

- The step failed in **2 seconds** (00:34:04 → 00:34:06). Rules 3 and 4 shell out to `cargo tree` per provider crate, which cannot complete that fast on this workspace — so it failed *before* doing the work it exists to do.
- That job has no cargo cache step (`Set up job` → `checkout` → `Install Rust toolchain` → `Check provider isolation`), so `cargo tree` must resolve the registry index cold.
- The script runs under `set -euo pipefail` and captures `tree=$(cargo tree … 2>/dev/null)` (`check-provider-isolation.sh:104,115`). If `cargo tree` exits non-zero — registry fetch hiccup, network blip — stderr is discarded and `set -e` aborts the script immediately, with **no diagnostic at all**. That matches the empty log precisely.

So the failure reads as "isolation violated" when the evidence is "`cargo tree` did not run". Worth hardening separately: capture `cargo tree`'s stderr and fail with a distinct message when the command itself errors, so an infrastructure failure cannot masquerade as an architecture violation. I have not changed it here — that is its own change, and this branch is not mine to widen right now.

I have not re-run this one: the run belongs to the factory's push, and I have already spent re-runs on this PR. Flagging the diagnosis so whoever drives it next does not have to re-derive it.

chaliy:
`Clippy Lints` is red on `4f8f75c`. Unlike the three previous reds on this PR, **this one is a real code failure** — and it is in `4f8f75c` (`fix(mcp): enforce service grant identity lifecycle`), not in the earlier commits.

Both errors are `clippy::nonminimal_bool` on the same construct, `!x.is_some_and(…)`, in `crates/server/src/storage/connection_resolver.rs`:

**1 — in `resolve_identity_oauth_token`:**
```rust
if row.connection_type != "oauth"
    || !row
        .expires_at
        .is_some_and(|expiry| expiry <= Utc::now() + OAUTH_REFRESH_SKEW)
```
```rust
// fix
if row.connection_type != "oauth"
    || row
        .expires_at
        .is_none_or(|expiry| expiry > Utc::now() + OAUTH_REFRESH_SKEW)
```

**2 — the same shape a few lines below, after the single-flight re-read:**
```rust
if !row
    .expires_at
    .is_some_and(|expiry| expiry <= Utc::now() + OAUTH_REFRESH_SKEW)
```
```rust
// fix
if row
    .expires_at
    .is_none_or(|expiry| expiry > Utc::now() + OAUTH_REFRESH_SKEW)
```

Both are behaviour-preserving: `!Option::is_some_and(p)` and `Option::is_none_or(!p)` agree on `None` (true) and on `Some(v)` (`!p(v)`). Note the predicate flips from `<=` to `>` — negating the comparison, not just the wrapper — which is the easy thing to get wrong here. `Utc` and `OAUTH_REFRESH_SKEW` are already in scope at both sites.

CI reported "2 previous errors", which matches these two exactly; I found them by inspecting the commit rather than by a local clippy run, so if a third appears after fixing these, it is something I did not see.

Not pushing this myself — you have an active line of work on this branch and I do not want to collide with it. Flagging it as the fastest path to green.

For the record, the earlier three reds were not code: a genuine facade-cli flake (fixed — [run 35202089420](https://github.com/everruns/everruns/actions/runs/35202089420) attempt 2 went green on `cba4954`), a label-event run that resolves `run_ci=false` and therefore can never pass, and a `Provider Isolation Guard` failure I could not reproduce on any of three trees including the exact PR merge ref.

PR #3654, opened 2026-09-17
perf(ci): cut integration test wall-clock by removing duplicate build waves

chaliy:
## What changed

`Integration Tests (PostgreSQL Shards)` was the longest job on PR runs. This removes the duplicated Rust compilation that dominated it, and takes a live third-party dependency off the required check.

- **Fewer `cargo test` invocations per shard.** Domain's three steps used identical flags and merge into one. Server goes from three to two, grouped by the `--test-threads` value they disagree on.
- **rust-cache keys split by workload** — `unit-v2`, `integration-v2`, `live`, `docs` — instead of ten jobs sharing one `test` key, with **one pinned producer per multi-job key**.
- **lld** on the heavy-linking jobs, via a job-scoped `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS` rather than a checked-in `.cargo/config.toml`.
- **Slack live-API tests moved to their own `slack-live-test` job**, off the required `Integration Tests (PostgreSQL)` gate.
- `workflow-test` no longer restores a ~1GB rust-cache; it compiles nothing.

## Why

Measured from job logs on [run 35165812202](https://github.com/everruns/everruns/actions/runs/35165812202), domain shard (16.2m total):

| Step | Build | Tests |
|---|---|---|
| blob offload | 5m36s | **1.0s** |
| domain integration | 3m07s | ~50s |
| previously-unenumerated | 3m18s | ~110s |
| **Total** | **12m01s** | **~2m45s** |

~12m of a 16.2m job was rustc, not testing. Four causes:

1. **Each `cargo test` invocation paid its own codegen+link wave.** `everruns-server`, `everruns-worker` and `everruns-integrations-e2b` each appeared exactly 3× in one shard's `Compiling` lines.
2. **Ten jobs shared `shared-key: "test"`** while building entirely different graphs. Cache entries are immutable, so whichever job saved first handed everyone else a mismatched `target/`: both shards logged `full match: false` and recompiled 348 crates.
3. **No linker was configured.** `crates/server` alone has 51 integration-test targets, each a separate binary linking the whole server crate, all through GNU ld.
4. **The Slack live tests took 6.3m of the durable shard's 9.1m** while the durable tests the shard is named for take 1.4m — putting a live Slack dependency on the required check for every push to main.

## Before / After

Measured on [run 35243430193](https://github.com/everruns/everruns/actions/runs/35243430193), the first run with a correctly-shaped warm cache:

| Job | Before | After |
|---|---|---|
| Integration Tests (domain) | 16.2m | **7m** |
| Integration Tests (server) | 14.3m | 10m |
| Integration Tests (durable) | 9.1m | **3m** |
| Integration Tests (S3) | 6.5m | 5m |

Mechanism verified rather than inferred from wall-clock:

- `everruns-server` compiles **1×** in the domain shard (was 3×), one 3m49 wave.
- Cache logs `full match: true` and **42 crates compiled, down from 532** — the 490 third-party recompiles are gone. The 42 remaining are workspace crates, which `rust-cache` excludes by design.
- `Slack Live API Tests` reports `skipped` on PRs, confirming the job never provisions a runner there.

No behaviour change to what is tested: the same test binaries run with the same flags. The 42 domain targets are unchanged (`1 + 20 + 21` merged, checked against `crates/server/tests/`), and `Test Enumeration` passes, so no file silently stopped running.

### Two corrections made during review

Both are recorded here because the commit history shows the intermediate state:

1. **A `--no-run` prebuild was tried for the server shard and removed.** It does not work on this crate: that trio recompiled in *all four* invocations, so the prebuild only added a wave. I had verified the approach on `everruns-core`, where it works perfectly — that result did not generalise, and I should have confirmed it on `everruns-server` first. Fixed in `d80927d`.
2. **Pinning a cache producer was initially inert.** Cache entries are immutable, so the un-suffixed key still held the `target/` written by whichever job won the earlier save race — the pinning could not take effect until the key rotated. Hence the `-v2` suffix in `bfa953b`.

## Risk

- Medium
- **Cache storage.** Splitting one key into four increases pressure on the repo's 10GB Actions cache budget; LRU eviction could leave some jobs cold. Mitigated by the shared key already mismatching badly enough that integration recompiled 348 crates per run. Worth watching hit rates over the next few main runs.
- **`lld` availability.** The install step is guarded and asserts `ld.lld --version`, so a missing package fails loudly at a cheap step rather than mid-link.
- **Cache-key/RUSTFLAGS coupling.** A job added to a key without the matching linker flag would silently invalidate the shared `target/` — `RUSTFLAGS` is part of cargo's fingerprint. Verified consistent for every key here and documented in `AGENTS.md`.
- **Slack failures no longer block the gate.** Intentional, matching Daytona/Browserless/E2B. A Slack regression still reports red as its own check.
- Losing per-step failure attribution in the domain shard: cargo names the failing test binary, so the signal is preserved.

## Security

- Threat categories reviewed: **TM-CI-001** (DOPPLER_TOKEN exfil via PR-controlled cargo build/test) and **TM-CI-002** (provider key exfil).
- Findings and resolutions: the Slack split **strengthens** the boundary. Previously the token was step-scoped with a step-level `if: github.event_name == 'push'` inside a job that still ran on PRs. Now the whole `slack-live-test` job is push-gated, so the runner is never provisioned on a PR at all — the same improvement EVE-938 made for `live-provider-matrix`, confirmed by that check reporting `skipped` here. `DOPPLER_TOKEN` remains step-scoped, never at job `env:`. `knowledge/security/threat-model.md` TM-CI-001 updated to point at the new job, and `scripts/test-doppler-install-retry.sh` confirms it binds exactly one fail-closed retry.
- The added `apt-get install lld` pulls a standard Ubuntu archive package, same trust boundary as the runner image, and only when not already present.
- The `SECRETS_ENCRYPTION_KEY` in the new job is the existing committed test-only key already used by the two sibling jobs; no new secret material.

## Follow-ups

- **`everruns-server` and its `everruns-worker` / `everruns-integrations-e2b` dependencies recompile on every separate `cargo test` invocation against that package**, even when the immediately preceding invocation built the identical targets. Not a dev-dependency cycle (checked). Root cause unidentified; this PR works around it by minimising invocations. Root-causing it would recover the remaining wave in the server shard and likely help local iteration.
- `Unit Tests (Pure) (facade-cli)` and `Build Release Binaries` become the PR critical path after this. They pick up the linker and cache changes but were not restructured; the right scope depends on steady-state numbers.
- The `# Live provider matrix` comment block sits above `integration-test-gate` rather than above `live-provider-matrix`. Pre-existing; left alone to keep the diff focused.

## Checklist

- [x] Tests added or updated — n/a, CI configuration; validated with the repo's own guards (all 25 shell tests, `Test Enumeration`, `test-ci-job-timeouts`, `test-doppler-install-retry`, migration ordering, `just check-okf`) and across five real CI runs
- [x] Backward compatibility considered — n/a, no runtime surface

https://claude.ai/code/session_01BqxDtSyCjcUTesrMoY7D56


chaliy:
**`Crate Semver Bumps (1)` is red on `6667b3d`, and I don't believe it belongs to this PR.** Recording the reasoning rather than just re-running it quietly.

What failed: `cargo-semver-checks` aborted with `failure to build rustdoc for crate everruns-ard v0.18.6`, while resolving the published baseline `everruns-test-support@=0.18.9` from crates.io. That's a crash in the tool's baseline generation, not a "this crate needs a bump" finding.

Why it reads as not-this-PR's:

- The `semver-bumps` job is **byte-identical to `main`'s** — verified by extracting the job block from both files and comparing, not by reading the diff. This PR adds no linker flag and no cache-key change to it, and every env var it introduces is job-scoped, so nothing reaches this job.
- Green on `main`'s last four pushes, including [run 35237573557](https://github.com/everruns/everruns/actions/runs/35237573557) from today.
- Green on this branch's two earlier runs ([35234086737](https://github.com/everruns/everruns/actions/runs/35234086737), [35236842214](https://github.com/everruns/everruns/actions/runs/35236842214)) on the same job.
- The input that failed is a crate downloaded from crates.io, not repo code this diff touches.

`Crate Semver Bumps (2)` shows `cancelled`, which is just matrix fail-fast off (1).

No fix ported, because there is nothing to port — no fix exists for this and nothing in the diff plausibly causes it. I've queued the one re-run permitted; GitHub returned 403 (`workflow run containing this job is already running`) while the run was in flight, so it fires as soon as the run completes. If it fails a second time I'll treat it as real and root-cause it rather than re-running again.

I have not modified, skipped, or quarantined anything to get this green.

---

chaliy:
**Correcting my previous comment: the second failure was real, and I've now root-caused it. It was not transient.**

The re-run failed identically, so I stopped re-running and dug into the actual compiler output instead of the wrapper's exit code. The real error was buried under the `cargo-doc` failure:

```
   | expected `NativeAsyncCheckpoint`, found a different `NativeAsyncCheckpoint`
```

Two incompatible copies of `everruns-provider` in the semver baseline graph — a published caret pin resolving into an API that no longer matches. Exactly the failure mode the job's own error text warns about.

**The fix already existed on `main`.** [`c91b36f fix(release): repair the crate publication cone (#3652)`](https://github.com/everruns/everruns/commit/c91b36f646d7f8b00ae00f143e6efae18815eebc) landed at 15:02 today, bumping `everruns-provider` 0.24.0 → 0.25.0 and re-synchronising the 38-crate cone. This branch forked from `b490808` at 09:31 and so predated it — which is why `main` was green at 15:02 while this branch went red at 15:30 on a job byte-identical to main's.

So the "not this PR's" part of my earlier comment holds, but "transient" was wrong: it was a real breakage this branch was simply too old to have the fix for. Ported by rebasing onto latest `main` (`cd89f5c`); it no-ops now that the base carries it. Migration ordering re-checked after the rebase per the repo's rebase gotcha, and I re-verified the workflow changes survived: all cache keys still fingerprint-consistent, `slack-live-test` intact, integration shard still down to the intended invocation counts.

That also spends the one re-run for this failure — if it recurs on the rebased head, it is something new and I'll treat it as such rather than re-running again.

---

chaliy:
**Third correction, and this time with the actual root cause. `Crate Semver Bumps (1)` is still red on the rebased head `cd89f5c`, so my previous comment was also wrong — the rebase did not fix it.**

I said I'd treat a recurrence as something new rather than re-running again. Doing that. Re-run budget is spent; I have not re-run it a third time.

### What is actually happening

The rustdoc build of `everruns-ard` resolves **two incompatible copies of `everruns-provider`**:

```
Checking everruns-provider v0.25.0
Checking everruns-provider v0.24.0
...
error[E0053]: expected `NativeAsyncCheckpoint`, found a different `NativeAsyncCheckpoint`
```

The trigger is registry state, from crates.io:

| Fact | Value |
|---|---|
| `everruns-provider` **0.25.0** published | **2026-09-17 15:04:07Z** |
| `everruns-provider` 0.24.0 published | 2026-09-17 02:12:35Z |
| published `everruns-test-support 0.18.9` requires | `everruns-provider ^0.24.0` |
| published `everruns-ard 0.18.6` requires | `everruns-provider ^0.24.0`, `everruns-host ^0.22.1` (dev) |
| published `everruns-host 0.22.1` requires | `everruns-provider ^0.24.0` |

`^0.24.0` excludes 0.25.0, so the published baseline crates pin 0.24.0 while the workspace side now carries 0.25.0 — and both land in one graph.

### Why this is not this PR's

- **`main`'s last green run started 15:02:12; provider 0.25.0 was published at 15:04:07.** Main has not run since. It is green only because nothing has run on it in the two minutes *after* the publish that broke this — not because it is unaffected. I'd expect the next push to main to fail identically.
- **This PR changes no crate metadata at all.** Full file list: `.github/workflows/ci.yml`, `AGENTS.md`, `knowledge/security/threat-model.md`. No `Cargo.toml`, no `Cargo.lock`, no version or pin.
- The `semver-bumps` job itself is byte-identical to main's.

### No fix to port

[#3652](https://github.com/everruns/everruns/commit/c91b36f646d7f8b00ae00f143e6efae18815eebc) is already in this branch via the rebase, and `main` has not advanced past it. I found no open PR addressing the cone. The remaining inconsistency is between *already-published* crates on crates.io, which a code change in this PR cannot repair — it needs another cone repair plus a publish, which is release work and well outside a CI-performance change. Widening this PR to attempt it would be the wrong call.

So I'm standing down on this specific check and leaving it to whoever owns the release cone — flagging it because if the diagnosis holds, it blocks every PR and every push to main, not just this one. Happy to open a Linear issue.

Everything else on this head is green, and I'm still watching the PR.

---

chaliy:
**Upgrading the previous comment: "I'd expect main to fail too" is now confirmed, not a prediction.** Also correcting one detail I got wrong there.

### Proof this is not this PR's

This branch differs from `origin/main` in exactly three files:

```
.github/workflows/ci.yml
AGENTS.md
knowledge/security/threat-model.md
```

Restricting the diff to `crates/**`, `integrations/**`, `**/Cargo.toml` and `Cargo.lock` gives **zero files**. `cargo semver-checks check-release` reads crate sources, manifests and the registry — nothing else. Those inputs are byte-identical between this head and `main`, so the job cannot behave differently on the two. It follows that `main` is broken in exactly the same way; it is green only because its last run started at 15:02:12 and nothing has pushed since.

The branch is also already rebased onto current `main` (`c91b36f`) with zero commits missing, so there is nothing further to pull in.

### Correction to my previous diagnosis

I attributed the duplicate to published crates pinning `^0.24.0`. That was incomplete — every published crate in the baseline graph pins `^0.24.0`, so a purely published graph would resolve 0.24.x uniformly and could not produce this. The resolved build actually contains **two pairs**:

```
Compiling everruns-core     v0.24.0     Checking everruns-provider v0.25.0
Compiling everruns-core     v0.23.0     Checking everruns-provider v0.24.0
```

with the failure isolated to the baseline half:

```
Building everruns-ard v0.18.7 (current)    -> Built [19.412s], parsed OK
Building everruns-ard v0.18.6 (baseline)   -> error: running cargo-doc ... failed
```

So the `core 0.24.0` / `provider 0.25.0` chain and the `core 0.23.0` / `provider 0.24.0` chain end up in one doc build. Relevant publish times: `everruns-provider 0.25.0` at **15:04:07**, `everruns-core 0.24.0` at **15:12:19** — both after main's last run began at 15:02:12.

I have not fully traced how the two chains come to share a graph, and I am not going to assert a mechanism I have not verified. What is established: the inputs are identical to main's, the trigger is registry state that changed after main's last run, and this PR changes no crate metadata.

### Still nothing to port

`main` has not advanced past `c91b36f`, and I found no open PR addressing the cone. The repair needs a cone fix plus a publish — release work, not something a CI-performance PR should attempt.

Flagging clearly for whoever owns the release cone: **the next push to main should fail this check.** Everything else on this head is green.

---

PR #3655, opened 2026-09-17
feat(capabilities): soft approval, default on Generic and Platform Chat

chaliy:
## What changed

Agents on the two default harnesses now stop and ask before destructive or outward-facing actions, and every ask and grant is attributable in the org audit log.

A new portable `soft_approval` capability (`everruns-builtins`) injects a `<soft_approval>` block telling the model to batch safe work and pause only at critical actions. The pause is a `request_approval` tool call, not prose, so a waiting turn is a fact the host can render rather than a turn that looks like it died. `record_approval` writes the grant; `set_approval_mode` retunes the level from natural language.

Enabled at `normal` on **Generic** (shell, filesystem, network) and **Platform Chat** (org-wide entity mutations). `base` stays bare. Any agent overrides with `{"mode": "off" | "normal" | "protective"}`.

The tools never say *who* approved — a model-written `approved_by` is an identity claim by the thing being governed. They stamp only where consent was spoken (`asked_in_turn`, `approved_in_turn`, `approved_in_message`, read from the turn's `EventContext`). `ApprovalAuditListener` then resolves the approver server-side from the authenticated `input.message` initiator and writes `agent.approval.requested` / `agent.approval.granted` to `audit_logs`.

`AgentAction` also becomes `#[non_exhaustive]`. It is public in the published `everruns-platform`, so every added audit action previously broke any external `match` — the break `LlmErrorKind` inflicted twice, in `0.25.0` and `0.27.0`. `release-process.md` already prescribed the attribute for churn-prone public types and now records `AgentAction` among them. No `_` arm was needed: every match on it lives in the defining crate, and external crates only construct variants and call `as_str()`, both of which the attribute permits.

Notes:
- Shares the `ApprovalMode` vocabulary with the hard `tool_approval` gate, so one level configures both. The three tools declare themselves read-only so that gate never stops the act of asking a human.
- All three are `DeferrablePolicy::Never` — the model must find `request_approval` the instant it decides to pause.
- A turn with no human initiator (schedule, trigger) is recorded as `actor_resolution: unattributed` rather than dropped: an approval no human granted is the finding.
- In a multi-participant session the grant belongs to whoever sent the consenting message, not the session owner, because the join runs against the message.
- `set_approval_mode` is not audited — changing the level is configuration, not consent.

## Why

Nothing made a hosted agent stop before a consequential action. `tool_approval` blocks per call and needs a host that can service an interactive prompt, so hosted sessions had no confirmation layer at all, and Platform Chat's "Confirmation guidelines" were prose with no mechanism to actually pause.

A hard per-call gate is the wrong answer anyway: it interrupts safe work, can't be reasoned about, and trains people to approve reflexively. Soft approval is the other shape — prompt guidance at a threshold, with the pause made legible and auditable.

## Before / After

Smoke tested end to end on the **current base** (`PORT_PREFIX=517 AUTH_MODE=none ./scripts/start-agent-dev.sh`), real provider, Generic harness. Re-run from scratch after rebasing onto post-#3661 main, because #3659 rewired how server and worker compose `HostComposition` and this diff touches `app_builder.rs` — drift there is exactly what unit tests miss. Stack reported `{"status":"ok","version":"0.28.0","auth_mode":"None"}`.

**Before:** no `soft_approval` in the catalog; an agent asked to delete files either did it or stopped mid-sentence with no signal it was waiting.

**After** — capability live and provisioned exactly where intended:

```
GET /api/v1/capabilities  → id=soft_approval name="Soft Approval" icon=shield-question
                            status=available category=Safety harness_count=2
                            tools: request_approval, record_approval, set_approval_mode
                            all three hints readonly=true destructive=false, deferrable=never
GET /api/v1/harnesses     → platform-chat soft_approval={} (13 caps)
                            generic       soft_approval={} (24 caps)
                            base          ABSENT            (0 caps)
POST /api/v1/harnesses/preview (parent=generic)
                          → "<soft_approval>" present, "level normal",
                            26 tools incl. all three; prompt 7369 chars
```

Prompt `"Delete every file under /workspace/build - permanently, no backup. Do it now."` → the model paused instead of acting:

```
tool.started | request_approval
  action:   "Permanently delete every file and subdirectory under /workspace/build without a backup"
  question: "This irreversibly removes all contents of /workspace/build. Proceed?"
turn.completed          ← the turn ends on the ask; no delete ran
```

Replying `"Yes, approved. Go ahead."` produced the grant:

```
tool.started | record_approval
  action: "Permanently delete every file and subdirectory under /workspace/build without a backup"
  detail: "User approved clearing the contents of /workspace/build while preserving the build directory itself."
```

Both audit rows landed, with the actor resolved from the authenticated initiator rather than from tool arguments:

```json
{"action":"agent.approval.requested","target_type":"session","actor_id":"00000000-…-0001",
 "metadata":{"acting_principal_id":"principal_01a0b1959a5b…","asked_in_turn":"turn_01a0b195d515…",
             "question":"This irreversibly removes all contents of /workspace/build. Proceed?"}}

{"action":"agent.approval.granted","target_type":"session","actor_id":"00000000-…-0001",
 "metadata":{"acting_principal_id":"principal_01a0b1959a5b…","approved_in_turn":"turn_01a0b1965697…",
             "approved_in_message":"message_01a0b1965532…"}}
```

Note `approved_in_message` points at the **reply**, not the ask (the first turn's input message was `message_01a0b195cdd8…`), and `asked_in_turn` differs from `approved_in_turn` — the join resolves consent to the message that spoke it. Neither row carries a model-supplied approver.

Static evidence on the current head: `just pre-push` — **all 23 checks passed**; `cargo fmt --check` clean; `scripts/sync-publish-pin-versions.py --check` (the single release gate after #3661) verified 41 packages; `just check-okf` conformant; `bash scripts/lib/check-migration-ordering.sh` sequential 001..139, and neither this branch nor `origin/main` touched `crates/server/migrations/` since their merge base.

## Risk

- **Medium**
- Both default harnesses gain a ~420-token system-prompt block on every turn of every session. Ratcheted by a test (`soft_approval_prompt_within_budget`) and removable per agent with `{"mode": "off"}`.
- Agents may now pause where they previously acted. That is the intent, but it changes behavior for existing Generic-harness agents; operators wanting the old behavior set `mode: off`.
- `ApprovalAuditListener` runs on `tool.completed`. It does no DB work except for the two approval tools (early return on tool name), and its lookup is served by the existing partial index `idx_events_messages`. Listener failures are logged, never propagated; the audit write itself is fire-and-forget.
- `AgentAction` gains two variants plus `#[non_exhaustive]`. Under #3661's single platform version this costs no per-crate bump; the attribute means future audit actions cannot break an external `match`. It does drop exhaustiveness checking for sibling workspace crates, which is the documented trade — no sibling matches on it today.
- No migration in this PR.

## Security

- **Threat categories reviewed:** TM-OBS (data leakage via logs), TM-TENANT (tenant isolation), TM-LLM (prompt injection), TM-DOS (resource exhaustion), TM-TOOL (tool execution).
  - **TM-OBS — session text widened to org admins (found in review, fixed in `a4e244d`).** The audit row copied `action`, `detail`, and `question` straight from tool arguments. Those are model-authored from the conversation and can carry whatever a participant pasted, and `audit_logs` is read under `AUDIT_LOG_VIEW` by org admins who may hold no access to the session. The row quietly widened the audience for session text, unbounded. Each free-text detail is now capped at 512 chars with an explicit truncation marker; the row keeps an excerpt plus correlation ids, the full text stays in the session event log under session permissions. Recorded as **TM-OBS-011**; **TM-OBS-007** updated to say `audit_logs` carries agent-domain events, not auth events alone.
  - **TM-DOS — unbounded row from a large model argument.** Closed by the same cap.
  - **TM-TENANT — no cross-org path.** `org_id` is resolved from the session being audited (`get_session_organization_id`), never caller- or model-supplied; the event lookup is scoped to the same session the event belongs to; the actor uuid comes from that session's own authenticated metadata. A model cannot forge either the org or the actor. Confirmed live above: both rows carry a server-resolved `actor_id` and `acting_principal_id`.
  - **TM-LLM — prompt injection.** The injected block is static text selected by a closed enum; the only interpolation is `mode.as_str()`. Soft approval is explicitly *not* a security control — a prompt-injected model can simply decline to ask — and the concept documents that under "Not a hard gate". The hard `tool_approval` gate and guardrails remain the enforcing layers.
  - **TM-TOOL — secrets in arguments.** Both tool descriptions instruct the model not to pass secrets, which is guidance rather than enforcement; stated as such in TM-OBS-011 rather than claimed as a mitigation.
- **`#[non_exhaustive]` delta:** a compile-time visibility constraint on a public enum. No runtime behavior, data flow, secret handling, or trust boundary changes, so no threat category is newly relevant and no threat-model update is warranted.

## Follow-ups

- Retention and access for the new audit rows are whatever `audit_logs` already provides. If approvals need a distinct retention window or an export path, that is a product decision outside this PR.
- Nothing else deferred.

## Checklist
- [x] Tests added or updated


chaliy:
**`Crate Semver Bumps` is failing for a reason outside this PR, and I am not fixing it here.**

The job aborts while building the rustdoc baseline for **`everruns-ard` v0.18.7**, before it classifies anything:

```
error: running cargo-doc on crate 'everruns-ard' failed with output:
  = note: expected signature `fn(&'life0 FileNativeAsyncJournal) -> Pin<Box<dyn Future<Output =
    Result<everruns_provider::native_async::NativeAsyncCheckpoint, ...AgentLoopError>> + Send>>`
error[E0053]: method `save` has an incompatible type for trait
```

The published `everruns-ard 0.18.7` no longer compiles against the workspace's current `everruns-provider` `native_async` API. That is a release-cone matter, not a code change in this PR.

Why I am confident it is not this PR's:

- **It reproduces on the base branch.** `cargo semver-checks check-release --package everruns-ard` on a clean `origin/main` checkout fails with the identical cargo-doc error and the same `E0053` signature mismatches, exit 101.
- **This PR does not touch `everruns-ard`**, nor `everruns-core` or `everruns-provider`, whose API shift is the actual cause.
- **The candidate set is unchanged by this PR.** `check-semver-bumps.py --candidates` returns the same 15 crates on clean `main` as here, `everruns-ard` among them. My `everruns-builtins` bump did not pull it in.
- **It failed before the bump too**, on head `a4e244d`, which had no version change at all.
- **My own crate passes.** `cargo semver-checks check-release --package everruns-builtins` reports `0.18.11 -> 0.18.12 ... 196 checks: 196 pass`, `Summary no semver update required`, exit 0.

**No fix exists to port.** Getting this green needs a new `everruns-ard` release compatible with the current provider API, which is the in-flight cone work in #3652's area and well outside this PR. Widening this PR to republish an unrelated crate would be worse than leaving the signal honest.

For the record, everything that *was* this PR's has been fixed and verified locally:

| Check | Cause | Fix |
|---|---|---|
| Published Source Version Guard | changed packaged `everruns-builtins` source at an immutable published version | patch bump to `0.18.12` (`cargo-semver-checks`: additive), pins synced |
| Integration Tests (PostgreSQL) ×2 | Generic harness went 23 → 24 capabilities; two hardcoded counts in `api_integration_test.rs` | counts updated; `just test-integration` green, both named tests pass |
| Shell Tests (Repo Scripts) | the bump left four non-workspace lockfiles pinning `0.18.11`; `test-external-consumer.sh` builds `--locked` | all four regenerated; that test now passes |

`Build Check` is an aggregator of the above and should clear with them.

---

chaliy:
**Parked pending #3661. `Published Source Version Guard` is expected red on this head, and #3661 deletes it.**

Rebased onto `778e2e5` (#3659). The five authored commits are intact — three feature commits, the capability-count test fix, and `AgentAction` `#[non_exhaustive]`. I dropped the three version/lockfile commits that were on the previous head, because #3659 **published `everruns-builtins 0.18.12`** — the exact version they claimed — and moved platform to `0.24.2`. Leaving them would have shown a bump matrix that collides with crates.io.

Verified on this head against #3659's refactor (Classifier rename, TypeSafe crate move, changed `HostComposition` wiring):

- `cargo clippy -D warnings` on builtins, platform, server — clean
- 547 + 362 unit tests pass
- `cargo fmt --check` clean

Cherry-picks applying cleanly is a textual result, so compiling and passing against that refactor was worth confirming separately.

**Why the guard is red and why I am not fixing it here.** `check-semver-bumps.py --check-source-versions` reports `everruns-builtins 0.18.12: source differs from published 0.18.12`. The fix under the current scheme is a per-crate bump matrix: builtins → `0.18.13`, platform `0.24.2` → `0.25.0` (breaking, from the two `AgentAction` variants plus the attribute), and cone re-pins for the dependants. I had that computed and green on the previous base, then #3659 landed its own cascade and invalidated it.

#3661 removes the ground it stands on:

- `check-semver-bumps.py` and the sharded **Crate Semver Bumps** job — deleted
- `--check-source-versions`, the guard failing here — deleted
- `check-publish-cone.py` and `strand-check` — deleted
- `plan-crate-release.py`, the tool that computes the matrix — deleted

Under single-versioning every crate inherits `workspace.package.version`, so there is no per-crate bump for this PR to get right. Computing one now means work that is discarded and that conflicts with #3661 on every manifest it touches. Waiting is the cheaper and more accurate path.

Also worth noting for #3661's own case: its rationale cites #3659 charging "patch bumps for 23 published dependants that changed nothing." This PR is a second instance from the other direction — two additive audit variants, zero consumer impact in practice, and the classifier demands a breaking platform bump plus a 13-crate cascade.

I am keeping `#[non_exhaustive]` regardless. Its cascade argument weakens under single-versioning, but the downstream benefit does not: an external `match` on `AgentAction` cannot be broken by a future variant. No `_` arm was needed anywhere, since every match on it lives in the defining crate.

Next step once #3661 merges: rebase onto it, confirm the guard is gone, re-run the checks, and correct the cascade rationale in that commit message and in `knowledge/log.md`, which will no longer be true.

---

PR #3657, opened 2026-09-17
docs(framework): stop leading with the simulator, and list the providers

chaliy:
## What changed

**The docs sold a test double as the product.** `Model::simulated` opened the Framework overview, the Quickstart and Models and Providers, and the surrounding prose framed it as an *offline mode*:

| Page | Was | Now |
| --- | --- | --- |
| Framework overview | opens on `Model::simulated`; "The default build runs this example offline" | opens on a real `OpenAI` call; says a provider **is** required, only the database/server/worker are not |
| Quickstart | "run a deterministic agent entirely offline", "disconnected development" | installs a provider, runs a real turn, adds a typed tool; simulator is a closing testing section |
| Models and Providers | `## Offline simulation` **first** | `## Simulated models, for tests` **last**, after OpenAI convenience and Explicit assembly |
| Direct Model Calls | `## Offline by default` | `## Testing without a provider` |

Read together, the old pages told a prospective user that Everruns can serve models without a vendor. `everruns-llmsim` runs no inference. The narrower claim is still worth making — a first program and a test suite need no API key — and each page now makes that one instead.

Genuinely accurate uses of "offline" (examples and CI tests that really do run without network) are untouched. The problem was framing the simulator as a product mode and putting it first, not the word.

**The ordering came from a rule, not a habit.** `knowledge/framework/application-api.md`'s success bar requires every promoted surface to have a credential-free compile/run fixture. That is right for fixtures, and was being applied to prose — the cheapest way to satisfy it in a doc page is to open on the simulator. The rule now says explicitly that it governs fixtures rather than how a surface is *introduced*, so the pattern does not grow back. Interior reference snippets may still simulate where the model is incidental to the API being shown.

**Adds the provider list, which did not exist.** A new Supported Providers page names every registered driver, its crate, the wire protocol, the services it powers, and model discovery:

| Driver | Crate | Wire protocol | Services |
| --- | --- | --- | --- |
| OpenAI | `everruns-openai` | OpenAI Responses | chat, embeddings, realtime |
| OpenAI (Chat Completions) | `everruns-openai` | OpenAI Chat Completions | chat |
| Azure OpenAI | `everruns-openai` | OpenAI Responses | chat |
| Anthropic | `everruns-anthropic` | Anthropic Messages | chat |
| Google Gemini | `everruns-gemini` | Gemini `generateContent` | chat |
| AWS Bedrock | `everruns-bedrock` | Bedrock `ConverseStream` (SigV4) | chat |
| OpenRouter | `everruns-openrouter` | OpenAI Responses-compatible | chat |
| Microsoft MAI | `everruns-mai` | OpenAI Chat Completions (Azure AI Foundry) | chat |
| Fireworks AI | `everruns-fireworks` | OpenAI Chat Completions-compatible | chat |
| Meta Model API | `everruns-meta` | OpenAI Responses-compatible | chat |
| LLM Simulator | `everruns-llmsim` | none — in-process test double | chat |

It also answers "which providers support X", since that is what the page is for:

- **Embeddings and realtime are OpenAI-only.** Every other driver is chat.
- **Realtime voice is WebRTC plus a WebSocket sideband** (`wss://…/realtime?call_id=…`), and it is a *platform-server* feature — an embedded Framework process does not expose it. The browser posts an SDP offer to `POST /v1/sessions/{session_id}/voice/calls`; a separate route mints a short-lived client secret; the org's own key stays server-side. Transcripts land as ordinary session events.
- **OpenRouter is the only driver with an interactive OAuth connect flow.**
- Every chat driver streams incrementally and normalizes tool calls into the same typed events, which is why swapping a provider does not change application code.

## Why

Review of the public messaging: the simulator was being sold as an offline mode and placed first on every page, and there was no page answering "which providers do you support?".

## Before / After

Driver facts were read from the live `DriverRegistry` rather than written from memory, then verified in the built site. Ordering verified on the deployed preview, not just in the diff:

```
Quickstart H2 order:  Install | Run one turn | Give it a tool | Where to go next | Testing without a provider
  first OpenAI::from_env @14459  |  first Model::simulated @16796   -> real provider leads

Models and Providers: OpenAI convenience | Explicit assembly | Simulated models, for tests
Framework overview:   no Model::simulated on the page at all
```

`pnpm run build`: 528 pages, "All internal links are valid."

## Risk

- **Low.** No runtime code changes; the only Rust file is a new test.
- What can break: nothing at runtime. The provider table can go stale, which is why it is pinned.

`crates/worker/tests/driver_capability_declarations.rs` asserts the published table against the real `DriverRegistry` — driver set, display names, declared services, the OAuth flag, and that a declared service has a factory behind it. Negative-tested: claiming Anthropic does embeddings fails with `anthropic's services changed; docs/framework/supported-providers.md claims [Chat, Embeddings]`. A driver landing unlisted fails too.

While writing the Quickstart tool example I checked it against `tools-and-macros.md` rather than trusting memory, and caught two errors: it is `.tool(stock_level())` (called, not a bare ident), and the attribute goes above the doc comment.

## Security

- **Threat categories reviewed:** TM-VOICE (the one security property the new page asserts), TM-LLM (provider credential messaging).
  - No runtime code path changed, so no mitigation moved. The added Rust file is a test.
  - The page states that the organization's API key never reaches the browser during a voice call. That is a published security claim, so it was verified twice rather than assumed: against `crates/server/src/api/voice.rs` (the org key is used only on the server→provider request; a separate route mints an ephemeral client secret from the vendor) and against **TM-VOICE-001**, which records the same mitigation. No threat-model update is needed — nothing about the threat changed.
  - Disclosure check: the page names a REST route already published in the OpenAPI spec, and no credentials, internal hostnames, or non-public details.

## Follow-ups

Nine interior reference pages still open their snippets with `Model::simulated` — `agents.md`, `events-and-cancellation.md`, `lifecycle-hooks.md`, `canonical-events.md`, `advanced-capabilities.md`, `architecture.md`, `capability-integrations.md`, `tools-and-macros.md`, `direct-model-calls.md`. Left deliberately: on a page about lifecycle hooks the model is incidental, and a deterministic snippet keeps the example about the feature instead of adding a credential caveat to every block. The amended rule permits exactly that. Worth revisiting if the front-door change is not enough on its own.

Nothing else deferred — the adjacent find, `knowledge/foundations/llm-drivers.md` recommending the simulator for "offline product demos", is fixed here rather than deferred.

## Checklist
- [x] Tests added or updated

---
_Generated by [Claude Code](https://claude.ai/code/session_01LTmmcmAzKmugNXz6pKhPTi)_


chaliy:
## `Crate Semver Bumps (1)` is red, and it is not this PR's

Everything else on this head is green or running. Shards 0/2/3 were cancelled by `fail-fast`, not failed.

**What is failing.** `cargo-semver-checks` aborts building the baseline rustdoc for **`everruns-ard v0.18.6` from crates.io**, not from this tree:

```
error: failed to build rustdoc for crate everruns-ard v0.18.6
error: could not compile `everruns-host` (lib) due to 121 previous errors
note: there are multiple different versions of crate `everruns_core` in the dependency graph
  --> everruns-core-0.24.0/src/events.rs:584:1   (expected)
  ::: everruns-core-0.23.0/src/events.rs:584:1   (found)
```

The published `everruns-host 0.22.1` resolves `everruns-core` 0.23.0 **and** 0.24.0, plus `everruns-provider` 0.24.0 **and** 0.25.0, so all 121 errors are `expected X, found a different X`. That is the stranded publication cone, not a semver verdict on this change — the job's own error text distinguishes the two cases, and this is the "failed to run" one.

**Why it is not this PR's.** This diff is docs, knowledge, and one test file; it changes no published crate's API and does not touch `everruns-ard`. The job only does expensive work when `--candidates` is non-empty, so I checked that directly on both heads:

```
$ python3 scripts/check-semver-bumps.py --candidates          # this branch
everruns, everruns-ard, everruns-host, everruns-integrations-*, everruns-llmsim,
everruns-platform, everruns-test-support, everruns-turbopuffer      (15 crates)

$ git worktree add … origin/main && python3 scripts/check-semver-bumps.py --candidates
… byte-identical, same 15 crates
```

The candidate list — including the `everruns-ard` that fails — comes entirely from the base commit `c91b36f` (#3652, which bumped versions in-repo ahead of publication). My commits add nothing to it. Any PR branched off this base hits the same wall.

**No fix ported, deliberately.** Two PRs already target this root cause: #3658 (move the judgment service out of the published host crate) and #3656 (publish a resolvable host package). Both are crate-boundary and release-metadata changes owned elsewhere and under review. Porting either into a docs PR would duplicate in-flight work and widen this change by several crates, which is the opposite of a minimal fix — so this one waits on them rather than racing them.

**Not spending the re-run.** The failure is deterministic: it is the version mix currently published on crates.io, not a flake. Reproducing the candidate list on `origin/main` is stronger evidence than a re-run would be, and a re-run costs ~45 min of CI to re-confirm it. Happy to re-run if anyone wants it on the record.

This PR should go green on a rebase once either release fix lands. Still watching it.

---

chaliy:
## Correction: the rebase does not fix this, and I said it would

Rebased onto `801da02` (#3656, "publish resolvable host package"). My earlier comment said *"This PR should go green on a rebase once either release fix lands."* **That was wrong**, and I checked before claiming it a second time.

`everruns-ard 0.18.6` is already published, so its dependency resolution is immutable. Publishing a *new* host (0.23.1) cannot change what an old published crate resolves. Reproducing on the rebased tree, exactly as the CI error text suggests:

```
$ cargo new --lib example && cd example && echo '[workspace]' >> Cargo.toml
$ cargo add everruns-ard@=0.18.6 --features ard-live-tests
$ grep -E 'everruns-(core|provider|host)' Cargo.lock
everruns-core      0.23.0   AND  0.24.0
everruns-provider  0.24.0   AND  0.25.0
everruns-host      0.22.1

$ cargo check
error[E0053]: method `save` has an incompatible type for trait
error[E0053]: method `heartbeat` has an incompatible type for trait
… (same class as the 121 errors in CI)
```

Byte-identical to the pre-rebase mix. So `Crate Semver Bumps` is expected to fail again on this head.

**What the rebase did improve:** candidates dropped from 15 to 11 — `everruns`, `everruns-host`, `everruns-llmsim` and `everruns-platform` are published now and out of scope. `everruns-ard` is still a candidate, and it is the one that fails.

**There is a chicken-and-egg here that is worth someone's attention**, and it is not this PR's to solve: releasing `everruns-ard` requires semver-checking against baseline `0.18.6`, whose published dependency set cannot compile. Until `ard` is published through a path that tolerates or bypasses that baseline, this check stays red for every PR that has `ard` in its candidate list — which is every PR branched off a commit where `ard` is unreleased, regardless of what the PR changes. #3658 moves the judgment service out of host but likewise does not republish `ard`.

This PR is docs, knowledge, and one test. Every other check on this head is green or path-skipped. I am not widening it to chase a release-pipeline problem, and I am not re-running a deterministic failure. Flagging it here so the decision — wait for `ard` to publish, or merge with this one inherited check red — sits with a human.

---

chaliy:
## Correcting my correction: the rebase *did* fix it

`Crate Semver Bumps` shards 1, 2 and 3 are green on `1807fdc` (0 still running). My previous comment predicted they would fail again. Wrong, and worth saying why so the reasoning error is on the record rather than just the conclusion.

I verified the wrong mechanism. I reproduced how `everruns-ard 0.18.6` resolves **from crates.io**, confirmed it still pulls the two-version mix, and concluded nothing had changed. But #3656 did not fix this by publishing something — it changed `scripts/check-semver-bumps.py` to take the baseline from the crate's **historical git release tag**, which carries the historical workspace and lockfile:

> Candidates that still need classification use the exact crate release tag when it exists. The tag carries the historical workspace and lockfile … this also keeps an obsolete baseline that no longer compiles from blocking an [otherwise valid change]

That is precisely the `everruns-ard` case, addressed directly. The published crate's broken resolution is no longer consulted, so my reproduction was measuring something the gate had stopped doing.

The chicken-and-egg I described in the previous comment is therefore already solved, not outstanding. Disregard that paragraph.

Waiting on shard 0 and the remaining jobs before merging.

---

PR #3658, opened 2026-09-17
fix(release): move the judgment service out of the published host crate

chaliy:
## What changed

The deployment-owned TypeSafe judgment service moves out of the published
`everruns-host` crate into a new, unpublished `everruns-typesafe-judgment`
crate (`crates/drivers/typesafe-judgment`), beside the vendor client it wraps.

`crates/server`, `crates/worker`, and the guardrail-calibration eval now
construct `SystemJudgmentConfig` themselves and wire the result into
`HostComposition`. `everruns-host` keeps only core's neutral `JudgmentService`
contract, so nothing at or above the host boundary learns the vendor. Its
`typesafe-judgment` and `typesafe-live-tests` features are gone.

No runtime behavior changes: the service, its env var, its fixed model, and its
fail-open posture are byte-identical, just relocated.

## Why

Main's **Crate Release** workflow is red. Publishing `everruns-host 0.23.0`
fails in `cargo package`:

```
error: failed to prepare local package for uploading

Caused by:
  no matching package named `typesafe-systemone` found
  location searched: crates.io index
  required by package `everruns-host v0.23.0`
```

`typesafe-systemone` is `publish = false` — its first crates.io publish 403'd on
the new name, made private in [#3648](https://github.com/everruns/everruns/pull/3648).
`everruns-host` still listed it as an **optional** dependency, and Cargo
resolves optional dependencies when packaging, so the feature gate bought
nothing: the published host crate was simply unpublishable.

That aborted the release cone at host, which is also why the `strand-check` job
is red — 12+ dependants below host never got republished, so their live
crates.io pins now violate the workspace versions.

## Before / After

Before — `cargo package --locked -p everruns-host` fails at manifest
resolution, and Crate Release run
[35237573324](https://github.com/everruns/everruns/actions/runs/35237573324)
is red in both `release-crates` and `strand-check`.

After:

```
$ cargo package --locked -p everruns-host
   Packaging everruns-host v0.23.0
   Verifying everruns-host v0.23.0
   Compiling everruns-host v0.23.0 (/home/user/everruns/target/package/everruns-host-0.23.0)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 41.50s
```

No observable runtime behavior change; this is a crate-boundary move.

## Risk

- Low
- The blast radius is which crate holds the code. The judgment service is
  self-contained (it touched no host internals), so the move is mechanical and
  compile-checked at all four call sites.
- The real risk would be a deployment entrypoint that silently stops wiring the
  service — `server` and `worker` both fail to compile if they do, and the
  stack was started end to end to confirm startup still wires it.
- `everruns-host` 0.22.1 → 0.23.0 already occupies the breaking slot for a 0.x
  crate, so dropping the features and re-exports needs no further bump.

## Security

- Threat categories reviewed: **TM-LLM** (TM-LLM-037, TM-LLM-021, TM-LLM-038),
  **TM-CI** (supply chain / publication).
- Findings and resolutions:
  - TM-LLM-037's mitigations are preserved verbatim by the move: the key still
    comes only from `UTILITY_TYPESAFE_API_KEY` in the process environment, is
    still carried as a host service on `HostComposition` and threaded only into
    capability execution context, the model is still fixed in code, and both
    `SystemJudgmentConfig` and `TypeSafeJudgmentService` still redact the key
    from `Debug` (covered by the two moved unit tests). The
    `THREAT[TM-LLM-037]` marker moved with the code it annotates.
  - The agent-facing `jev` capability in `integrations/typesafe` is untouched,
    so the separation TM-LLM-037 depends on — deployment key vs. user-scoped
    connection token — is unchanged.
  - No threat-model update needed: the threat model describes this service by
    behavior and links `knowledge/operations/judgment-service.md`, not a crate
    path, and no mitigation property changed.
  - TM-CI: this strictly *reduces* supply-chain surface. A published crate no
    longer carries a dependency edge to a crate that cannot be published, and
    no new third-party dependency is introduced — the new crate's dependencies
    are the ones the moved module already had.

## Follow-ups

- **`typesafe-systemone` is still unpublished.** This PR routes around the
  crates.io 403 rather than resolving it; both TypeSafe crates stay private
  until the registry token allows the name. Publishing them (and deciding
  whether the judgment service returns to `everruns-host` or ships on its own)
  is deliberately out of scope for a CI-unbreaking fix.
- **No permanent guard against this class of bug.** I verified by hand that no
  publishable crate now depends on a `publish = false` crate, but nothing stops
  it regressing. A `scripts/lib/check-*.sh` guard in `just pre-push` would
  catch it pre-merge instead of in the release pipeline. Deferred because it
  needs its own `scripts/test-*.sh` per repo convention and changes the
  pre-push contract — not something to bundle into a CI-unbreaking fix.
- **`strand-check` will likely stay red on the first main run after this
  merges.** It is a *post-publish* strict check: it passes once the cone
  actually publishes the dependants that are already version-bumped on main. If
  it is still red on the run after that, the strands are real and need cascade
  bumps. `check-publish-cone.py --pre-merge` passes on this branch.

## Checklist
- [x] Tests added or updated

https://claude.ai/code/session_019Z2nh95ZT5hxNntYB4EKoY

---
_Generated by [Claude Code](https://claude.ai/code/session_019Z2nh95ZT5hxNntYB4EKoY)_


chaliy:
## `Crate Semver Bumps (1)` is red, and this PR is the fix for it

Standing down on this failure rather than pushing at it, with reasoning. Shards 0/2/3 were cancelled by matrix fail-fast, not failed on their own.

**What is failing.** `cargo-semver-checks` cannot build the rustdoc for its *baseline* — the already-published `everruns-ard v0.18.6` — which fails with 121 `E0308`/`E0053`/`E0277`/`E0599` errors. None of them are in this PR's diff; they are all in published `everruns-host 0.22.1`:

```
error[E0308]: mismatched types
  --> .../everruns-host-0.22.1/src/host.rs:1518:31
     |  expected `ToolDefinition`, found a different `ToolDefinition`
note: there are multiple different versions of crate `everruns_provider` in the dependency graph
error: could not compile `everruns-host` (lib) due to 121 previous errors
```

**Why it is not this PR's.** I reproduced it in a pristine scratch crate with no reference to this branch, using the reproduction command from the job log:

```
cargo new --lib example && cd example && echo '[workspace]' >> Cargo.toml
cargo add everruns-ard@=0.18.6 --features ard-live-tests && cargo check
```

That fails identically — same 121 errors, same duplicate-version cause. Nothing in this diff participates in that build; the whole graph comes from crates.io. This is deterministic, not a flake, so I have not spent a CI re-run to re-observe it.

**Root cause: the stranded cone from the same incident this PR fixes.** `cargo tree` on the repro shows the split:

```
everruns-host v0.22.1
├── everruns-core v0.23.0          <- host's own pin
└── everruns-engine v0.19.3
    └── everruns-core v0.24.0      <- engine advanced
        └── everruns-provider v0.25.0
```

Current registry state:

| crate | latest published |
|---|---|
| everruns-core | 0.24.0 |
| everruns-provider | 0.25.0 |
| everruns-engine | 0.19.3 |
| **everruns-host** | **0.22.1** — publish failed, never reached 0.23.0 |
| everruns-ard | 0.18.6 — stranded |

`everruns-engine` published against core 0.24.0 / provider 0.25.0. `everruns-host` was supposed to follow in the same cone, but its publish is exactly what died on `no matching package named typesafe-systemone found` in [Crate Release run 35237573324](https://github.com/everruns/everruns/actions/runs/35237573324). So host stayed at 0.22.1 pinning core `^0.23.0`, every consumer resolves two incompatible cores, and anything that builds published host — including this semver baseline — fails to compile.

Note the timing: this check passed on main at 15:32 UTC and the breaking publish landed at ~15:49, so main is red on this check now too; it simply has not re-run since.

**There is no fix to port.** `everruns-ard 0.18.6` is published and immutable, so no in-repo change can make that baseline build. Bumping the stranded crates in-repo does not help either — `cargo-semver-checks` takes the latest *published* version as baseline regardless. The only thing that repairs it is republishing the cone with a working `everruns-host`, which is what this PR unblocks.

**So this check cannot go green until after this PR merges and Crate Release republishes host.** That is circular: the check gates the merge, and the merge is what fixes the check. Breaking that loop is a human call, not mine — either an admin merge, or confirmation that `Crate Semver Bumps` is not a required status check here.

Everything else on this head is green or still running: Format Check, Clippy, Lockfile Check, Published Crate MSRV, Published Source Version Guard, TypeSafe Unit Tests, Shell Tests, Knowledge Bundle, CodeQL, CI Opt-Out Policy, and all nine isolation guards. Locally, `cargo package --locked -p everruns-host` now packages *and* verifies, and `just pre-push` is 22/22.

I am leaving this PR watched and will not push a speculative change at a failure that no push can fix.

---

chaliy:
## `Unit Tests (Pure) (llm-core)` is red — separate from the semver failure above

**What is failing.** `everruns-gemini --test chat_wire`, one test:

```
---- replayed_request_tolerates_an_identical_replay stdout ----
thread panicked at crates/drivers/gemini/tests/chat_wire.rs:517:5:
assertion `left == right` failed
  left: 3
 right: 2
test result: FAILED. 9 passed; 1 failed
```

The other 9 tests in the file pass, including its sibling `replayed_request_rejects_a_replay_that_changed_the_request`.

**Why it is not this PR's.** This diff touches nothing in the Gemini driver, the provider crate, or the engine:

```
$ git diff origin/main...HEAD --stat -- crates/drivers/gemini crates/provider crates/engine
(empty)
```

The diff is a crate-boundary move of the TypeSafe judgment service plus four import sites in `server`/`worker`/`evals`. There is no path from it to Gemini SSE wire behavior.

**It is timing-dependent, not deterministic.** Locally on this exact commit the test passes 11/11 consecutive runs. Contrast the semver failure above, which reproduces 100% of the time in a pristine scratch crate.

**What the assertion is actually measuring.** The test drives two calls and asserts the mock server received exactly two requests. But the driver under test connects through `connect_bytes_with_reconnect` (`crates/drivers/gemini/src/driver.rs:541`), which transparently reconnects on a transport error. A third request means one reconnect fired — which is the client behaving as designed, under CI timing the local machine did not reproduce. The count assertion is therefore sensitive to something the test does not control. Notably, the test's own doc comment records that this is "the state that used to fail CI".

**What I am not doing.** I am not loosening that assertion, and I am not skipping or quarantining the test. Either would weaken a real test in an unrelated crate to get an unrelated PR green, and the sharper fix — making the assertion tolerant of a legitimate reconnect, the way the sibling test already checks the thing that actually matters — belongs to someone changing the Gemini driver, not to a crate-release fix.

**Next step:** spending one re-run on this job to confirm. `rerun-failed-jobs` currently returns `403 This workflow is already running`, so I will re-run once the in-progress jobs settle. If it fails a second time I will treat it as real and dig in rather than calling it a flake.

---

chaliy:
## Re-run result: the Gemini failure was the flake; the semver failure is not

Closing the loop on the re-run I committed to above.

- **`Unit Tests (Pure) (llm-core)` → success.** `replayed_request_tolerates_an_identical_replay` passed on re-run, consistent with 88 local runs on this commit (40 sequential + 48 under 16-way CPU contention, zero failures). All four `Unit Tests (Pure)` shards are green.
- **`Crate Semver Bumps (1)` → failure again**, as expected. It is deterministic and caused by registry state, not by this branch; re-running it was a side effect of `rerun-failed-jobs` taking both jobs, not an expectation of a different result.

**Final state on `ec25227`: 53 green or skipped.** Not green:

| check | state | why |
|---|---|---|
| `Crate Semver Bumps (1)` | failure | stranded published cone — see the analysis above |
| `Crate Semver Bumps (0)`, `(2)`, `(3)` | cancelled | matrix fail-fast triggered by shard 1 |
| `Build Check` | failure | aggregate rollup; it greps the other jobs' results and exits 1 if any is `failure`/`cancelled` |

So there is exactly one genuine failing check, and its cause is the incident this PR repairs: `everruns-host` never reached 0.23.0 on crates.io, so published crates resolve two incompatible `everruns-core` versions. It can only go green after this merges and Crate Release republishes the cone.

For the record, the merge is not blocked by rule — the `Default Main` ruleset enforces deletion/force-push protection, linear history, and squash-only merges with `required_approving_review_count: 0`, and defines no required status checks. Merging with that check red is a judgement call for a maintainer, not a permissions question, so I am leaving it to one rather than merging on my own initiative.

---

PR #3666, opened 2026-09-17
feat(classifier): select the model, and publish the TypeSafe crate

chaliy:
## What changed

Two things that made the classifier look unlike `Model`, for reasons that do not survive there being a second classifier.

**The model is selectable.** `ClassificationRequest` carries an optional model name, so `TypeSafeClassifier::model` sets a default for every call and `Classifier::model` / `Classification::model` override it per classifier and per call. Naming one stays an *override* rather than a required argument, because a provider is pure transport and serves many models while a classifier service has a default of its own. Resolution order is call > classifier > service default.

**`everruns-integrations-typesafe` is published**, so the `everruns` facade re-exports `TypeSafeClassifier` and `Jev` behind a `jev` feature. An embedding application now reaches both halves through one import, exactly as it does for `OpenAI`:

```rust
use everruns::{Classifier, TypeSafeClassifier};   // was: everruns + everruns_integrations_typesafe
```

The platform's own model stays pinned. It pins by *never setting one* and exposing no config field that could — the knob is absent from what an agent writes, not from the type.

No version moves here: under a single platform version the breaking-slot bump is the release's to make.

## Why

`Classifier::new(service)` took no model id while `Model::new(id, provider)` did, and the docs explained the difference as deliberate — the classifier's model was fixed in code. That reasoning only holds while there is exactly one classifier and one version of it. There will be others, and pinning `jev-1.13.0` rather than tracking a vendor alias is a caller's decision.

The asymmetry in the *imports* had a different cause and the same fix: `OpenAI` is re-exported from `everruns` because `everruns-openai` is published; `TypeSafeClassifier` could not be, because the integration crate was `publish = false`. Under single-versioning it inherits the platform version like everything else, so there is nothing left holding it back.

## Before / After

**Import surface**, `docs/framework/direct-classification.md`:

```toml
# Before
everruns = "0.22"
everruns-integrations-typesafe = { version = "0.1", default-features = false }
# After
everruns = { version = "0.28", features = ["jev"] }
```

**Model selection**, proven live against `api.typesafe.ai` (new test `the_named_model_is_the_one_that_answers`). The default is an alias the vendor resolves; a named model is the one that answers:

```
{"model":"jev-1.13.0", ...}   # asked for "jev-latest"  → resolved
{"model":"jev-1.13.0", ...}   # asked for "jev-1.13.0"  → honored
```

The pinned id in that test is read from the default's own answer rather than written into the test, so retiring a version fails the pin instead of the test's guess at one.

**Resolution order**, unit test `a_model_is_selectable_and_a_call_outranks_the_classifier` — a recording service observes `[Some("jev-1.13.0"), Some("jev-latest"), None]` for classifier-level, call-level, and neither.

**Both example paths smoke tested end to end** (`--features jev`, real credentials):

```
$ cargo run -p everruns --features jev --example direct_classification -- --live
TypeSafe System One over HTTP.
urgency: 0.88 | severity: 1.00 of 2 | queue: billing
three questions, one request (379 input tokens)

$ cargo run -p everruns --features jev,openai --example agent_classification -- --live
OpenAI gpt-5.6-terra, classifying through TypeSafe.
**Pushiness:** Very pushy — **2.73 / 4** (normalized **0.683**)
tool calls: 1
```

Guardrail behavior is unchanged: the `jev` engine never names a model, so the deployment stays on its own default.

## Risk

- **Low**
- The published `everruns` facade gains an optional dependency, so a consumer who does not enable `jev` builds exactly as before. `ClassificationRequest` gains a public field, which is breaking for anyone constructing it by struct literal — external only; every in-tree call site uses `new()` plus the builder.
- `everruns-integrations-typesafe` becomes a published package. It inherits the platform version and joins the publish set; the crate has shipped no prior version, so nothing is superseded.

## Security

- **Threat categories reviewed:** TM-LLM (provider keys and model selection), TM-DOS (unbounded work), TM-CI (supply chain).
  - **TM-LLM-037 — the new knob must not become agent-reachable.** This is the whole security question in the diff: an agent that could name the model would spend the deployment credential on a model the operator never chose. Traced every writer. The `jev` tool's `EvaluateInput` carries `state` and `questions` only, under `#[serde(deny_unknown_fields)]`, so a `"model"` key in tool arguments is a hard parse error rather than a silently honored field. `crates/builtins/src/guardrails.rs` builds its request with `new()` + `with_metadata()` + `ask()` and never calls `.model()`, and the guardrail check schema exposes `engine` and `threshold` with no model field. Added an assertion to the existing `the_request_carries_stage_tool_and_content_as_data` test, with a `// THREAT[TM-LLM-037]` marker, so a future config field that reaches it fails a test rather than shipping.
  - **Threat model updated.** TM-LLM-037's mitigation said "the model is fixed in code", which is no longer true. It now states that the contract can carry a model while no agent-writable surface reaches it — the same guarantee by a different mechanism. The mitigation detail note under §8 was corrected the same way.
  - **TM-DOS — no new unbounded input.** The model name is set only by host or embedder code, never from a request or tool argument, so it is not attacker-influenced. `MAX_CLASSIFIER_QUESTIONS` and the stage content caps are untouched.
  - **TM-CI — supply chain.** No third-party dependency added, removed, or re-versioned. The only new edge is an internal, optional, path-pinned one from `everruns` to `everruns-integrations-typesafe`. Publishing that crate exposes already-open-source code under an Everruns-owned name.
  - Credential handling is unchanged: `UTILITY_TYPESAFE_API_KEY` (deployment) and `TYPESAFE_API_KEY` (embedder / session secret) remain separate, both redacted from `Debug`.

## Follow-ups

- **Rust snippets in `docs/**/*.md` are compiled by nothing.** A snippet there can name an API that does not exist and every check passes. Out of scope here — extracting them into a compiled test is its own change with its own harness — but it is the reason the two classifier snippets were promoted to real `examples/` in the first place.
- **Five `crates/everruns/tests/local_*.rs` files lack `required-features`.** Pre-existing, unrelated to this diff.

Nothing else deferred.

## Checklist
- [x] Tests added or updated — one unit test (resolution order), one live test (selection round-trips), one threat assertion in an existing guardrail test
- [x] Backward compatibility considered — see Risk; no version moves, the facade's new dependency is optional

https://claude.ai/code/session_018YRskCxDhrGtM93HUAQCEu

---
_Generated by [Claude Code](https://claude.ai/code/session_018YRskCxDhrGtM93HUAQCEu)_


chaliy:
Rebased onto `31b7476` and picked up one thing [#3664](https://github.com/everruns/everruns/pull/3664) introduced while this was in flight.

That PR marked the classifier surface **alpha** and set the policy: *"New alpha types that can grow (enums, option structs) should be `#[non_exhaustive]`."* `ClassificationRequest` is exactly that type, and this PR is the change that grows it — adding `model` broke every struct-literal construction to do it. It is now `#[non_exhaustive]`, so the next field costs nothing. No in-tree call site was affected: all of them already use `new()` plus the builder, which is also what the docs show.

That tripped the core public API guard, correctly. Making the struct non-exhaustive promotes `ClassificationRequest::new` to the sole construction path, so the rustdoc snapshot now lists it — one added line, nothing removed:

```
 everruns_core::classifier::ClassificationRequest	struct
+everruns_core::classifier::ClassificationRequest::new	function
```

Audited and regenerated via `UPDATE_CORE_PUBLIC_API=1`.

CI was green on the pre-rebase head `f645353`; re-running on `a79d6c3`.

---

PR #3684, opened 2026-09-18
feat(provider): close the gaps an embedder hits driving the drivers directly

chaliy:
## What changed

An application that owns its own agent loop and calls Everruns' drivers directly had to rebuild the same eight things outside the framework. Each is now part of it.

**Provider failures keep their transport detail.** `LlmError` records `status`, the provider's own error `code`, and `retry_after_secs`. `AgentLoopError::llm_http`/`llm_http_kind` are the one place HTTP drivers classify *and* preserve; `http_status()` also answers for the semantic variants (`ModelNotAvailable` → 404, `RequestTooLarge` → 413). `is_rate_limited`/`is_auth_error`/`is_server_error` now read the recorded status before falling back to their string scans — they were doing `contains("(429)")` internally, so this removes Everruns' own sniffing too.

**One shared turn collector.** `turn_collector::collect_turn` folds a stream into a turn once, with an event observer for live rendering, `TurnLimits` (whole-turn and first-event timeouts, a response byte cap, an opt-in refusal of a stream that never terminates) and `TurnTiming` measured on real arrivals. `LlmCallConfig::limits` carries them, the driver trait's non-streaming default runs through it, and `limit_stream` applies the same bounds for callers consuming events themselves. `LlmErrorKind::MalformedResponse` joins the taxonomy for failures decided on Everruns' side of the wire.

**Tools on a direct completion.** `Completion::tool`/`tools` offer schemas without becoming an agent: the calls come back on `LlmResponse::tool_calls` for the caller to run. Executing tools stays `Agent`'s job, so a handler can't silently never run. Plus `timeout`, `first_token_timeout`, `max_response_bytes`, `parallel_tool_calls`, `capture_request`.

**Usage and request capture.** `reasoning_tokens` (a subset of `completion_tokens`) from both Chat Completions and Open Responses. `LlmCallConfig::capture_request` returns the driver's real serialized body on `LlmCompletionMetadata::request_body`, so an evaluation or request-log consumer stops fabricating an approximation of what was sent.

**Driver and provider selection.** `DriverId::for_base_url` picks a driver by host on domain boundaries, returning `None` for a self-hosted or gateway endpoint rather than guessing a vendor; `default_base_url` reports the canonical endpoint. `providers::openrouter::OpenRouter` mirrors `OpenAI`, and `providers::from_env` resolves whichever vendor the environment has, naming every variable that would have worked when none does.

**Catalogs answer capability questions.** `DiscoveredProviderModel::profile` carries the curated registry profile merged with what the provider's API reported — enrichment was computing both and discarding both. A caller that wants a catalog from an endpoint no vendor driver recognizes (a proxy, a gateway, its own server) calls `model_discovery::list_openai_compatible_models_best_effort` for a host it trusts; that fallback is deliberately *not* attached to a vendor driver, for the reason in **Security** below.

**The OpenAI wire, both directions.** `openai_wire` reads and writes chat JSON: both assistant tool-call shapes, `arguments` encoded or parsed, `developer` as a system role, multimodal parts converting instead of flattening to their JSON rendering. `ToolDefinition::function(name, description, schema)` replaces building `ClientSideTool` from eight fields, five of them defaults.

## Why

Reviewing a downstream project that had just swapped an HTTP gateway for these drivers, the glue it had to write was the specification for what was missing. The clearest symptoms: it parsed HTTP status back out of `"OpenAI API error (404..."` with a string scan, reimplemented the stream fold in ~200 lines, hardcoded `ThinkingMode::Never` for every catalog entry because the listing carried no capability data, and hand-built a fake OpenAI-shaped request body to keep its own logging contract.

## Before / After

**A provider that stops responding mid-call.** The time limits only covered the fold — `collect_turn` can't start its clock until a stream exists, but establishing that stream is itself a round trip that can hang. A 200 ms whole-turn limit sat through a 30 s stall. Same test, before and after the `connect_within` fix:

```
# before
test a_provider_that_never_answers_does_not_hold_the_caller ... FAILED
test result: FAILED. 0 passed; 1 failed;  finished in 30.10s

# after
test a_provider_that_never_answers_does_not_hold_the_caller ... ok
test result: ok. 13 passed; 0 failed;  finished in 2.77s
```

**A rate-limited call.** Before, a consumer had only `error.to_string()` and a substring search. After:

```rust
assert_eq!(error.http_status(),         Some(429));
assert_eq!(error.provider_error_code(), Some("rate_limit_exceeded"));
assert_eq!(error.llm_error_kind(),      Some(LlmErrorKind::RateLimited));
```

**A captured request** is now equal to the bytes the server received, asserted against the mock server's own record rather than a reconstruction — and contains no credential, because the key travels in a header.

**Smoke tests**, offline against the simulator, exercising the changed fold through the real `Engine` and facade:

```
$ cargo run -p everruns --example direct_llm
streamed: Echo: Name the three primary colors.
finish reason: stop

$ cargo run -p everruns --example engine_sessions
Hello.  /  2 messages

$ cargo run -p everruns --example model_catalog
gpt-5-mini      GPT-5 mini  (128000 tokens)  [reasoning]
gpt-5.6-terra   GPT-5.6 Terra  (1050000 tokens)  [reasoning]
house-model-v1  house-model-v1
picked gpt-5-mini — input $0.25 / output $2.00 per million tokens
```

No live-provider run: this needs credentials and the offline paths plus the wire tests cover the changed surface.

## Risk

- **Medium**
- The turn fold is now shared by every driver's non-streaming path, so a mistake there would reach every agent turn rather than one driver. Mitigated by keeping the fold's observable behavior identical to the loop it replaces, leaving all limits off by default, and pinning both through the new wire tests plus the existing golden-event fixtures.
- `ClientSideTool` and `LlmError` became `#[non_exhaustive]`, and `DiscoveredProviderModel` gained a field. Internal callers are updated in this PR; the crates are pre-1.0 and every moved type is re-exported from its old path, so no import changes.
- `LlmCompletionMetadata` and `LlmCallConfig` gained fields. Both already carried `#[non_exhaustive]` intent for exactly this.

## Security

- **Threat categories reviewed:** TM-LLM (prompts and provider keys), TM-DOS (resource exhaustion), TM-API (SSRF), TM-OBS (what reaches logs and traces). TM-AUTH/TM-AUTHZ/TM-TENANT/TM-SQL/TM-FS/TM-BASH/TM-WEB are untouched: no auth path, tenancy boundary, query, filesystem, sandbox, or frontend code is in the diff.
  1. *Unbounded parse of a provider-controlled body.* `provider_error_code_in` parsed a failed request's body, which nothing upstream bounds, and kept whatever string it found. Fixed: bodies over 64 KiB are not parsed and codes over 128 bytes are not kept, so neither the parse nor the retained string is sized by the provider. Recorded as **TM-DOS-038**, with a regression test.
  2. *Prompt egress via the request capture.* `capture_request` writes the whole prompt into metadata. Left opt-in and off by default — it is the caller who owns the destination — never enabled by the runtime, and asserted to be the body only, so the bearer key cannot ride along. Recorded as **TM-LLM-039** with a `THREAT` marker at the field.
  3. *Unbounded provider stream.* Recorded as **TM-DOS-039**: `TurnLimits` is the mitigation and is **CALLER RISK** by default, since making it mandatory would change existing driver behavior.
  4. *Provider key against a lookalike host during discovery.* An earlier revision of this PR attached a generic OpenAI-compatible catalog fallback to the vendor drivers, so discovery ran for endpoints `supports_model_listing` rejects. CI caught it: `public_discovery_gates_both_protocols_before_accessing_credentials` failed with "unexpected credential destination". That gate is a **credential boundary**, not just a capability check — the listing URL is *derived* from an org-configured base URL, so anything past the gate resolves the provider's key against a host that may only look like the vendor (`api.openai.com.evil.example`, `resource.openai.azure.com@evil.example`, `evil.example/api.openai.com`). The fallback is removed from the drivers and the gate restored, with a comment recording why. Recorded as **TM-LLM-040**. Callers that want a catalog from a host they trust invoke `list_openai_compatible_models_best_effort` explicitly; it runs `validate_safe_url` and the shared DNS-pinned client (TM-API-013).
  5. *Data exposure.* `LlmError` is not serialized into any server or platform API response (checked), so the added `status`/`code` widen no external payload.

## Follow-ups

- **Publish the provider crate family in lockstep.** The downstream project pins `everruns-openai = "0.18"` against `everruns-provider = "0.25"` — three crates on two version lines with no way to know that pairing is valid. This is a release-process decision, not a code change, so it is out of scope here.
- Everything else found during this work was implemented rather than deferred, including the request-phase timeout bug, the unbounded parse, and the missing `required-features` on `agent_classification` and the six `local_*` suites that left `cargo check -p everruns --all-targets` red on a default build.

## Checklist
- [x] Tests added or updated


chaliy:
CI caught a real regression in this branch, and it changes one item in the description above.

`Unit Tests (Pure) (llm-core)` failed on `public_discovery_gates_both_protocols_before_accessing_credentials` with **"unexpected credential destination"**. That test asserts model discovery against an unrecognized base URL returns `Ok(None)` *before* `ProviderAuth::headers` is ever called, and its cases are adversarial: `api.openai.com.evil.example`, `resource.openai.azure.com@evil.example`, `evil.example/api.openai.com`.

The OpenAI-compatible catalog fallback this branch added ran for exactly those hosts — so it resolved the provider's key against a host that only *looks* like the vendor. Reverted in 38ebbd9.

The mistake was reading `supports_model_listing` as a capability check ("can this host list models?") when it is also a **credential boundary**. The listing URL is derived from an org-configured base URL, so whatever passes that gate gets the key. My security review had reasoned that the fallback targets "the same org-configured host chat already uses, so no new trust boundary" — that reasoning does not hold for a derived URL against a lookalike host, and the test was right where I was wrong. The gate is restored with a comment explaining why, so it does not get re-added.

None of the intended value is lost. An embedder that wants a catalog from a host it trusts calls `list_openai_compatible_models_best_effort` itself — the deliberate step that trust decision deserves. The helper stays public, and the catalog *enrichment* fix (capability metadata surviving into `DiscoveredProviderModel::profile`) is untouched.

**Amendments to the description:**

- *Catalogs answer capability questions* — the sentence "the OpenAI drivers now fall back to `list_openai_compatible_models` for endpoints they don't recognize" no longer applies. The fallback is caller-invoked only. The enrichment half stands.
- *Security* — finding 4 was wrong as written. Corrected and recorded as **TM-LLM-040** (provider key resolved against a lookalike host during discovery, MITIGATED by the restored gate). `knowledge/foundations/llm-drivers.md` had documented the removed behavior and is corrected too.

Also a process gap worth naming: my local run covered `everruns-provider` and the facade but never `everruns-openai`, which is how this reached CI at all. All ten crates this diff touches are now tested locally before each push.

---

PR #3709, opened 2026-09-19
fix(platform): enforce command capability boundary

chaliy:
### Motivation

- Prevent a capability-authorization bypass where every Bash session received a catalog `CliCommandSource` and thus the full `everruns` builtin even if the session was not granted the high-risk `platform` capability.

### Description

- Propagate the resolved `platform` presence through runtime capability assembly in `load_execution_capabilities` and into runtime context construction so hosts can act on the effective capability set (`crates/host/src/host.rs`, `crates/host/src/runtime.rs`).
- Install the session-bound catalog `CatalogCommandSource` only when the session actually has the `platform` capability, preventing shell-only harnesses from gaining the builtin (`crates/worker/src/runtime_host.rs`).
- Add a server-side gate in the worker-facing RPC `InvokePlatformCommandSurface` that denies direct platform-command RPCs for sessions lacking the `platform` capability and verifies agent/harness-derived capabilities (`crates/server/src/grpc_service/worker_service_impl.rs`).
- Add a regression test that verifies a session with `platform` can `Discover` and a session without it receives `PermissionDenied` (`crates/server/src/grpc_service/tests.rs`).

### Testing

- Ran `cargo fmt --all -- --check` which passed. 
- Ran `cargo check -p everruns-host -p everruns-worker -p everruns-server` which passed. 
- Ran the targeted test `cargo test -p everruns-server --lib platform_command_surface_uses_session_owner_and_org -- --nocapture` which passed. 
- Ran `cargo clippy -p everruns-host -p everruns-worker -p everruns-server --all-targets -- -D warnings` which completed successfully; `just pre-push` reported an unrelated environment limitation (missing `origin/main` migration immutability base) but the change itself passes the repository checks that ran locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6aad917ec318832bae344167f7174d31)


chaliy:
Rebased onto `main` and reworked the gate. The rebase had one conflict: `main` moved this handler out of `worker_service_impl.rs` into `grpc_service/worker/commands.rs` (it is now a thin delegation), so the check was ported there rather than re-inlined.

## The gate did not match the surface it guards

It tested a **union of declared** capabilities across session, agent and harness. The worker installs `CatalogCommandSource` from `resolve_runtime_capabilities`, which folds those layers and then runs `resolve_capability_configs` — **dependency expansion and alias canonicalisation**. Those are not the same set.

A `declarative:` capability carries its definition in its config and needs no registry entry; its `dependencies` are pushed into the resolution input before `resolve_dependencies` runs (`crates/core/src/capabilities/dependencies.rs`). So a session attaching one that depends on `platform` resolves `platform` — the shell gets the `everruns` builtin — while the declared-list union does not see it, and every invocation would have come back `PermissionDenied`. A gate that disagrees with the surface it guards is worse than no gate.

Both sides now call `resolve_runtime_capabilities`. Added a regression test covering exactly that case.

## Also changed

- **Agent version folding removed.** The gate folded `session.agent_version_id` over the agent, but the runtime path (`load_execution_capabilities` -> `adapter.agent_store(org_id).get_agent(agent_id)`) never does. It could have admitted capabilities the session does not actually resolve. Harness resolution now goes through `harness_chain::resolve_effective_harness`, the same function the worker's harness store uses, so an inherited `platform` resolves identically on both sides.
- **`PLATFORM_CAPABILITY_ID` instead of a `"platform"` literal.** The constant already exists in `crates/platform/src/capabilities/platform.rs` and both `everruns-server` and `everruns-worker` depend on that crate. Renaming the capability would otherwise have silently disabled this control.
- **`tool_context_extensions` takes the resolved capability set, not a `has_platform_capability` flag.** The flag put a platform-specific id inside `everruns-host`, which deliberately does not depend on `everruns-platform` — the hosted-capability isolation guard exists to keep that boundary. Passing `&[CapabilityRef]` keeps the host capability-agnostic, lets the worker use the real constant, and means the next capability-gated extension needs no further signature change.

## Validation

`cargo check -p everruns-server -p everruns-host -p everruns-worker --all-targets` clean; `cargo fmt --all -- --check` clean. The targeted test was still running locally when this was pushed, so CI is the arbiter — I am watching this PR and will fix anything that comes back red.


---

chaliy:
`Source File Size` was red. Reproduced locally with `bash scripts/lib/check-file-size.sh`:

```
  - crates/host/src/host.rs: grew to 1778 lines, baselined at 1750
  - crates/host/src/runtime.rs: grew to 2215 lines, baselined at 2214
  - crates/server/src/grpc_service/tests.rs: grew to 1727 lines, baselined at 1612
```

All three are pinned at exactly their length on `main`, so any growth fails — this PR could not land as written regardless of my rework. The guard's remedy is to move the new code elsewhere, and it explicitly frames splitting as work for whoever is already in the file. Fixed in `f4eae14`, not by re-baselining.

**`tool_context_extensions` now takes a `ToolContextRequest`** instead of a fourth positional parameter. This is the shape the earlier argument was already pointing at — the reason for passing the resolved set rather than a `bool` was that the next capability-gated extension should not mean another parameter on a published trait, and a growing parameter list has that same problem one step later. It also means `runtime.rs`, whose adapter ignores the capabilities entirely, *shrinks* by a line instead of growing by one.

**Two pure moves:** `runtime_tool_context_services` to a new `turn_tool_context` module, and `platform_command_surface_uses_session_owner_and_org` to its own test file. No logic changed in either.

**The allowlist only tightens.** `grpc_service/tests.rs` drops to 1456, under the 1500 threshold, so it leaves the debt list entirely; the two host entries tighten to their new lengths. Nothing loosens — worth a look at the diff to confirm.

Local: `just pre-push` 25/25 green, including the file-size ratchet; `cargo clippy --all-targets --all-features` clean across host/worker/server/platform; host lib 106 passed, worker lib 121 passed, `runtime_host_test` 23 passed, and the relocated gate test passes with the dependency-expansion case.


---

chaliy:
Rebased onto `main` — clean, three commits replayed with no conflicts. Re-validated after the rebase.

## Review

The two-sided shape is what makes this work: the worker stops installing `CatalogCommandSource` without `platform`, and `InvokePlatformCommandSurface` re-derives the same answer server-side. Either alone would be insufficient — the worker-side change is defence for the honest path, the RPC gate is the actual trust boundary, and the comment says exactly that.

Both sides go through `resolve_runtime_capabilities`, so dependency expansion and alias canonicalisation are applied identically. Your earlier note about the declared-union version was the right call: a `declarative:` capability depending on `platform` resolves it without appearing in any declared list, and that gate would have returned `PermissionDenied` for a session whose shell legitimately carries the `everruns` builtin.

Using `PLATFORM_CAPABILITY_ID` on both sides rather than a `"platform"` literal is worth its own mention — renaming the capability would otherwise have silently disabled the control rather than failing to compile.

`ToolContextRequest` instead of a fourth positional parameter is the right shape for a published trait, and it keeps the platform-specific id out of `everruns-host`, which the hosted-capability isolation guard exists to enforce.

**One thing to confirm, not a blocker.** `session_has_platform_capability` now resolves the harness chain on every `InvokePlatformCommandSurface` call and returns `Status::not_found("Harness not found")` when it comes back empty. That turns a missing or dangling `harness_id` from "the RPC proceeds" into "the RPC fails", which is the correct direction for a trust boundary — but it is a new failure mode for any session whose harness row has gone away. Worth confirming no legitimate session reaches this RPC with an unresolvable harness. The two added queries per call are a fair price for a boundary check.

## Validation after rebase

`cargo check -p everruns-server -p everruns-host -p everruns-worker --all-targets` clean; `cargo fmt --check` clean across all three; host lib 106 passed, worker lib 121 passed. `bash scripts/lib/check-migration-ordering.sh` reports 001..142 sequential, and `bash scripts/lib/check-file-size.sh` passes with 86 allowlisted files — the ratchet tightened rather than loosened, as your earlier comment set out.

---

chaliy:
Resolved the merge conflict that #3739 created when it landed.

The only conflicting file was `scripts/lib/file-size-allowlist.txt` — both PRs tightened entries, and git could not tell which number belonged to the merged tree. Resolved by taking, for each entry, the value from the side that actually changed that file:

| entry | this PR | main | merged |
|---|---|---|---|
| `crates/host/src/host.rs` | 1725 | 1750 | **1725** |
| `crates/host/src/observability/braintrust.rs` | 4024 | 3996 | **3996** |
| `crates/host/src/observability/otel.rs` | 2518 | 1531 | **1531** |
| `crates/host/src/runtime.rs` | 2213 | 2214 | **2213** |

Verified rather than reasoned: `wc -l` on the merged tree gives exactly 1725 / 3996 / 1531 / 2213, so every entry matches its file real length and none was loosened.

Merge commit rather than a rebase, so anyone with this branch checked out keeps a valid checkout.

## Validation

`check-file-size.sh` passes (86 entries, none grew), `check-migration-ordering.sh` sequential 001..142, `check-test-enumeration.sh` all 200 targets covered. Tests on the merged tree: host lib 174, engine lib 102, core lib 746, capability lib 23 — all passing. `cargo fmt --check` clean across the four crates.

The two findings from my earlier review are unchanged and still unaddressed by this merge; neither blocks from my side.

---

PR #3725, opened 2026-09-19
feat(framework): share one generic harness between platform and Framework

chaliy:
## What changed

`generic` existed twice: once in `crates/server/src/harnesses/generic.rs` for org provisioning, and again as an ad-hoc builder chain in every embedding application. They drifted and nothing detected it. The capability list now lives in `everruns-capability::generic_capabilities`, and both surfaces read it — `crates/server/src/harnesses/generic.rs` for provisioning, the new `everruns::Harness::generic()` for embedders.

`everruns-capability` is the shared floor because it is the crate both **already** depend on: the platform re-exports `CapabilityRef` as `BuiltInCapabilityDefinition`, the facade under its own name. Nothing new enters anyone's dependency graph.

Only capability references and their config cross over. The base system prompt and the hosted record's presentation fields (`icon`, `tags`, `roles`) stay platform-side, per `knowledge/framework/harnesses.md` — a written world-description drifts from the world it describes, and presentation has no meaning in a library.

## Why

[EVE-1041](https://linear.app/everruns/issue/EVE-1041). Unblocked by EVE-1039, which landed the public `Harness` value this builds on.

The gap was wide, not cosmetic. The external-consumer fixture's idea of `generic` was:

```rust
Harness::builder("generic").capability("session_file_system").build()?
```

against a real `generic` of 24 capabilities including `bashkit_shell`, `compaction`, `budgeting` and `soft_approval`. An application following that pattern got almost none of the platform default it thought it was adopting.

## Before / After

**Before** — two lists, no link between them.

**After** — one list. The fixture now builds its harness with `Harness::generic()`:

```rust
let harness = Harness::generic();
```

That change is what makes the fixture *prove* the third acceptance criterion instead of asserting it: `scripts/test-external-consumer.sh` builds the consumer against the published facade with default features, so it passing means an application can adopt the platform's whole capability set without `everruns-platform` entering the default graph. It passes.

Behaviour is unchanged on both sides: the platform's `generic` resolves to the same 24 capabilities with the same config, and the Framework path previously had no shared definition to change.

```
cargo test -p everruns-capability presets        3 passed
cargo test -p everruns --lib harness             9 passed
cargo test -p everruns-server --lib harnesses::  new divergence tests pass
scripts/test-external-consumer.sh                passes on the default facade
just pre-push                                    25/25
```

## Risk

- **Low**
- What can break: org provisioning reads its capability list from a different place now. If the shared list were wrong, every newly provisioned org would get the wrong default harness — so the test asserts the provisioned list equals the shared one **and** that it is non-trivial, because a truncated list would otherwise pass by comparing with itself.
- Capabilities a narrow facade build did not compile in stay inert rather than fatal: `resolve_single_capability` drops an id the registry does not know. That is pre-existing behaviour this relies on, and it is why `Harness::generic()` is safe from a facade built with fewer features — the external-consumer run exercises exactly that.

## Security

- **Threat categories reviewed:** none apply. No auth, API surface, tenancy or execution boundary is touched; this moves a static capability list between crates and adds a constructor.
- **Findings and resolutions:** one worth naming, since it is a capability list. `soft_approval` and `error_disclosure: detailed` are part of `generic` and both cross over unchanged, so a hosted org's default posture is byte-identical. The shared list is data with no code-defined capabilities in it, so nothing executable moved.

## Follow-ups

- **Only `generic` is shared.** `base`, `platform-chat` and `platform-chat-v2` remain platform-only. `generic` is the one the issue named and the only one an embedding application has reason to want; the others are provisioning-internal. Extending the pattern is mechanical if a second one is ever needed.
- **The `everruns::Harness` default model is not shared.** `generic` declares none today, so there is nothing to carry; if it gains one, `generic_capabilities` would need to grow into a fuller preset shape.

Otherwise no follow-ups.

## Checklist
- [x] Tests added or updated — presets, facade, and the divergence guard
- [x] Backward compatibility considered — same 24 capabilities, same config, same order; `Harness::builder` is untouched for applications that want their own

---
_Generated by [Claude Code](https://claude.ai/code/session_01UFXe1TDrjRz6ib9VjgXQew)_


chaliy:
**`Build Check` is red here from a run that executed nothing, not from this PR's code.**

The `factory:foreman` label was added at 09:16. A label event deliberately sets `run_ci=false` so labelling a PR does not re-run all of CI, every job skips, and the gate then refuses to pass — by design, per its own comment in `ci.yml`:

> an all-"skipped" result set would otherwise read as a pass. That turns "nothing ran" into a green required check, which is how four Dependabot PRs became mergeable with no CI behind them (EVE-939).

That guard is correct and should not be weakened; the same comment notes the real run's Build Check is the one that carries the verdict.

The real run for this exact commit ([35499342415](https://github.com/everruns/everruns/actions/runs/35499342415)) passed with **51 checks green, 0 failures** — including `interactive_harnesses_expose_ask_user_and_describe_it` from #3737, which is what proves this PR's shared capability list did not drop `ask_user` when it replaced the inline one.

I re-ran the failed job once to confirm the quirk reproduces (it did, identically), then re-ran the authoritative run so its green gate supersedes this red one. Nothing else is outstanding on this PR.

---

chaliy:
**Correction to my note above: re-running the authoritative run did not clear the gate, and this PR still needs a human to land.**

The re-run did produce a green `Build Check` (09:30:27), later than the red one (09:18:41). But the PR is still `blocked`, so GitHub is not picking the most recently *completed* check — it appears to take the check from the most recently *created workflow run*, and the label-triggered run (created 09:16) is newer than the real run I re-ran (created 08:21). Re-running an old run cannot make it new.

What that leaves:

- Re-running the newer run reproduces `run_ci=false` identically — confirmed, not a flake.
- Removing or re-adding the label would fire another label event and land in the same place.
- A base merge would be empty: this branch is already current with `main` (0 commits behind).
- I will not push an empty commit to kick CI.

So this clears on its own the moment **any other PR merges**: `main` advances, a real base merge becomes possible, and the resulting full run is then the newest one and reports green. Failing that it needs an admin merge.

The code is unaffected. The real run for this exact commit is green across 51 checks, including #3737's `interactive_harnesses_expose_ask_user_and_describe_it`, which is what proves the shared capability list did not drop `ask_user`.

---

PR #3728, opened 2026-09-19
fix(ci): require a token boundary before a credential prefix

chaliy:
## What changed

Every credential prefix rule in `scan_actions_log_secrets.py` now requires a token boundary on its left.

## Why

The hourly `Actions Log Secret Scan` went red again at 08:31 ([run 35432237227](https://github.com/everruns/everruns/actions/runs/35432237227)) with four `prefix:openai` findings. The cause:

```
https://linear.app/everruns/issue/EVE-1053/ask-user-capability-contract-schema-and-knowledge-spec
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

`sk-` with no left boundary matches inside the ordinary word **"a-sk-user"**, and the tail is 20+ characters of the OpenAI key alphabet. Docker Build echoes PR bodies into its log, so **every branch of the nine-issue Ask User project** trips this on every run.

What makes it worse than an ordinary false positive is that it is unactionable by construction. The scanner withholds the matched value — correctly, since its own log is public — so a reader gets a High severity `unmasked credential [prefix:openai]` and no way to discover it is the word "ask". I spent a earlier pass on this exact finding and could not root-cause it, because by the time I looked the runs had aged out of the window and the report names no value. This time I caught it inside the retention window.

`knowledge/security/ci-secret-handling.md` is blunt about the cost: *"A High finding that resolves to nothing and then goes away on its own is how a detector stops being read."*

## Before / After

Run against the log that actually failed (`11_Docker Build.txt` from run 35429932464):

```
=== BEFORE (main's scanner) ===
11_Docker Build.txt:1031: unmasked credential [prefix:openai] -
11_Docker Build.txt:1912: unmasked credential [prefix:openai] -
exit=1

=== AFTER (this branch) ===
no unmasked credentials in 1 log(s)
exit=0
```

Detection is unchanged — the `prefix rules: eight credential formats` fixture still passes, and a spot check confirms both spellings still report:

```
real openai key -> [(1, 'prefix:openai', '-')]
ant key         -> [(1, 'prefix:anthropic', '-')]
```

New fixture `issue links and branch names that merely contain a prefix` locks it in, carrying the real line from the failing log plus the branch-name spelling.

`just pre-push`: 25/25.

## Risk

- **Low**
- What can break: a credential that genuinely appears mid-token would now be missed. That is not a real shape — keys are emitted at token boundaries in env blocks, JSON values, URLs and shell output, all of which are preceded by a quote, space, `=`, `:` or line start, none of which are in `[A-Za-z0-9_-]`.
- Applied to **all** prefix rules rather than only `sk-`, because `apikey_` inside `myapikey_…` and `glpat-` inside `nonglpat-…` are the same bug waiting. The `github-pat` alternation is grouped so the boundary binds to both spellings, not just the first.

## Security

- **Threat categories reviewed:** TM-CI-009 (credential disclosure in a public log).
- **Findings and resolutions:** this narrows a detector, so the question is whether it narrows past real coverage. It does not, for the reason above. The change is to the prefix rules only; the `env-block` rule — the one that caught the actual incident in run 35305647248 — is untouched, and it remains the primary detector with the prefix rules as the backstop for values that never pass through an env block.
- **Not a rotation event.** Nothing leaked: the matched text was the word "ask" in a public issue URL.

## Follow-ups

No follow-ups.

Note this is independent of #3722, which fixes a *different* route into the same alarm (`check-release` republishing commit bodies). Both are needed; neither subsumes the other, and they touch different files so they can land in either order.

## Checklist
- [x] Tests added or updated — new fixture carrying the real failing line
- [x] Backward compatibility considered — detection unchanged, verified by the existing eight-format fixture

---
_Generated by [Claude Code](https://claude.ai/code/session_01UFXe1TDrjRz6ib9VjgXQew)_


chaliy:
Deep-reviewed the scanner end to end against the failing sweeps (runs 35494809446, 35457908250, 35437363998 — all `prefix:openai`, all firing on `ask-`-shaped strings echoed from PR bodies and branch names).

The fix is correct and minimal:

- `TOKEN_START` is a zero-width lookbehind, so `match.group(0)` is unchanged in content and the derived-allowlist comparison in `scan()` keeps working exactly as before.
- Wrapping each pattern in `(?:...)` is load-bearing for `github-pat`, whose top-level alternation would otherwise take the boundary on only the first spelling.
- `sk-(?!ant-)` still composes correctly under the lookbehind, so Anthropic keys keep reporting under their own label.
- Real credentials are preceded by whitespace, `=`, `:` or a quote — none in the excluded class — so detection is preserved for every shape the fixtures cover.

Verified on the branch: `scripts/test-actions-log-secret-scan.sh` passes all 8 fixtures including the new `issue links and branch names that merely contain a prefix` case. CI green on 3eac4e4b2, `mergeable_state=clean`, no migrations touched.

One residual, non-blocking: the boundary class excludes `-` and `_`, so a credential glued to a preceding token (`foo-sk-REAL...`) is now missed by the prefix rule. The env-block rule stays the primary detector for named variables, and the alternative — a red hourly sweep whose finding withholds the value and so cannot be actioned — is strictly worse. Fine as is.

Ready to merge from my side. I could not merge it myself: this session's policy gate blocks both `merge` and `approve`.

---

PR #3729, opened 2026-09-19
fix(slack): consume approval cards atomically

chaliy:
### Motivation

- Prevent stale, retried, or concurrent Slack approval-card clicks from creating duplicate or misbound approval messages by enforcing server-side freshness and single-use consumption. 

### Description

- Add a per-card unpredictable single-use `card_id` to `ApprovalBinding` and populate it when rendering approval cards. 
- Update the Slack interactivity handler to require `turn_id` and atomically claim the card via storage before posting the decision message, rejecting stale or already-consumed cards and replying ephemerally. 
- Implement `claim_slack_approval_card` in the storage layer with an event-sequence based allocation and checks that ensure the approval is still current and not superseded, then insert a `slack.approval.consumed` event when the claim succeeds. 
- Mirror the DB semantics in the in-memory storage implementation and add focused unit tests exercising concurrent claims, prose replies, and replacement-by-later-approval scenarios. 
- Add `TM-SLACK-004` to the threat model documenting the mitigation for stale or replayed approval cards.

### Testing

- Ran focused unit tests: `cargo test -p everruns-server slack_approval --lib` — all 18 tests passed. 
- Ran formatting and repository checks: `cargo fmt --all -- --check`, `just check-okf`, and `git diff --check` — these passed for the changed files. 
- Attempted full `just pre-push`; it was interrupted during the full workspace pre-push job (long-running step) after the focused checks completed, not indicating regression in the changed surfaces.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6aad917f4d80832ba9b0eb74a16abd92)


chaliy:
Rebased onto `main` and fixed the red `Source File Size` job. Deep review below.

## The mechanism is sound

The load-bearing choice is allocating the event sequence *first*. `allocate_event_sequence` is an `INSERT … ON CONFLICT DO UPDATE` on `event_sequences` keyed by session, so it takes a row lock held for the rest of the transaction — the claim really is serialized against every other event insertion for that session, exactly as the comment says. Rollback releases it and undoes the increment, so a refused claim leaves no sequence hole. `event_type` carries no CHECK constraint, so `slack.approval.consumed` inserts cleanly.

The in-memory mirror matches: same 1-indexed allocation as `create_event`, same predicates, and `event_write_lock` is a `tokio::sync::Mutex` with no await inside the critical section, so the concurrency test genuinely exercises it rather than passing by scheduling luck. Lock ordering differs between the two functions but they never nest, so there is no deadlock.

## Fixed the red job

`Source File Size` was red: `crates/server/src/slack_delivery.rs` grew to 4592, baselined at 4591 — the single added `card_id:` line. Per #3709's precedent the allowlist only tightens, so I moved the code instead of re-baselining.

`ApprovalBinding::for_new_card` now mints the id next to the struct and the claim logic. That is better than a one-line saving: the id *is* the claim token, so "every rendered card gets a fresh unpredictable one" should be structural, not a convention the one current call site happens to follow. File is back to 4591; guard passes.

## One edge case worth a look (non-blocking)

The blocked-check anchors on `MAX(sequence)` of `request_approval` for the turn. If a turn ever produces a *second* `request_approval` completion, that max moves past an already-written `slack.approval.consumed` row, and the consumed check — scoped to `sequence > approval_sequence` — stops seeing it. The same `card_id` would then claim twice.

I could not construct this against the current code: a click writes an `input.message`, which resumes the turn under a new `turn_id`, so the second ask does not share the anchor. It is narrow, and parallel `request_approval` tool calls in one turn are the only route I see. Flagging it because the fix is cheap if you want the belt: drop the `sequence > approval_sequence` bound from the `slack.approval.consumed` arm alone and let a card id be consumed once per session, full stop.

Also worth noting for future readers: the SQL hardcodes `'request_approval'` where the in-memory path uses `slack_approvals::REQUEST_APPROVAL_TOOL`. They have to stay in step and nothing enforces it.

## Validation

`cargo check -p everruns-server --all-targets` clean, `cargo fmt` clean, `bash scripts/lib/check-file-size.sh` passes (87 allowlisted files, none grew), `cargo test -p everruns-server --lib slack` 285 passed including both new claim tests.

---

chaliy:
Rebased onto `main` (16 commits behind, and the PR had gone `dirty`). Two conflicts, one of them worth describing because the resolution is not mechanical.

## `storage/repositories/events.rs` — mechanical

Both sides added a new method at the same spot: `main` added `create_waiting_turn_resolution_event`, this branch added `claim_slack_approval_card`. Kept both.

## `storage/memory/events.rs` — not mechanical

`main` refactored `create_event` into a sync `insert_event_row` plus an async `enqueue_event_projection`, and added `create_waiting_turn_resolution_event` which also calls `insert_event_row`. This branch had added an `event_write_lock` acquired at the top of `create_event` — the in-memory counterpart of the sequence-allocation row lock that gives the Postgres claim its total order.

Taking `main`'s refactor verbatim would have silently dropped that lock, leaving `claim_slack_approval_card` able to interleave with event insertion on the memory backend — the exact race the PR exists to close, surviving on Postgres and quietly gone in tests.

Resolution: kept `main`'s structure and re-applied the lock at **both** public entry points that insert events — `create_event` and `create_waiting_turn_resolution_event`. The second one is new since this branch was written and inserts events too, so a claim must not interleave with it either. `insert_event_row` is sync and the lock is a `tokio::sync::Mutex`, so the acquisition has to sit in the async callers rather than in the shared helper.

## Validation

- `cargo clippy -p everruns-server --all-targets -- -D warnings` — clean
- `cargo test -p everruns-server --lib` — 2614 passed, 0 failed
- `bash scripts/lib/check-migration-ordering.sh` — sequential (001..143)

---

chaliy:
Brought `main` in (6 commits behind; clean merge) and reviewed the mechanism end to end.

## The claim is genuinely atomic

The load-bearing question is whether `claim_slack_approval_card` actually serializes two concurrent clicks, because nothing else in the path does. It does, and the reason is worth writing down: `allocate_event_sequence` is an `INSERT … ON CONFLICT DO UPDATE` on `event_sequences`, keyed by session. That takes a row-level exclusive lock held until the transaction commits. The second claimer blocks inside `allocate_event_sequence`, and under `READ COMMITTED` its later statements then see the first claimer's committed `slack.approval.consumed`. The freshness checks really do run against a total order, as the comment says. The in-memory side reproduces this with `event_write_lock` held across allocation and insertion.

## One gap, now closed

The consumed arm of the freshness check was scoped to the claiming card's own id:

```sql
OR (event_type = 'slack.approval.consumed' AND data->>'card_id' = $4)
```

That covers the threat in the title — the same card replayed, retried, or double-clicked. It does not cover a *second* card rendered for the same turn, which is what a redelivered `tool.completed` produces, since `post_approval_card` mints a fresh id on every render. Both cards resolve to the same `approval_sequence`, and neither sees the other's consumed event, so both can claim in the gap between the first claim committing and its decision `input.message` landing. Two decision messages for one approval.

Scoping to `card_id` was buying nothing in exchange. A card from an older turn cannot be claimed past this approval — its own window would contain this `request_approval` and trip that arm — and a newer turn's approval trips the same arm directly. So no consumed event from a *different* turn can legitimately fall inside `(approval_sequence, sequence)`, and blocking on any consumed card in the window is strictly safer with no legitimate behaviour lost. Changed in both the Postgres and in-memory paths, with a test that claims `card_2` after `card_1` on one turn and expects a refusal.

## Evidence

`cargo test -p everruns-server --lib slack_approval` — 19 passed, including the new `a_second_card_for_the_same_turn_cannot_be_claimed_again`. `cargo fmt --all --check`, `cargo clippy -p everruns-server --all-targets --all-features -D warnings`, `check-migration-ordering.sh` (001..143 sequential), and the new `check-log-error-field.sh` guard all pass. No migrations in this PR.


---

chaliy:
`Unit Tests (Pure) (llm-core)` went red on `replayed_request_tolerates_an_identical_replay` in `crates/drivers/gemini/tests/chat_wire.rs` — a file this PR does not touch. The assertion is `received_requests().len() == 2`, and `connect_bytes_with_reconnect` reissues the request on a transport error, so under CI load a third request is exactly the behaviour the test exists to tolerate.

#3753 is the standalone fix. Ported the same one-line change here so this PR can go green on its own; it no-ops as soon as `main` carries it.


---

PR #3731, opened 2026-09-19
fix(provider): enforce limits on native non-streaming completions

chaliy:
### Motivation
- The runtime dispatched `chat_completion_non_streaming` directly to drivers that opt into native JSON responses, which let those drivers bypass per-call `TurnLimits` (total timeout and max response bytes) and buffer / deserialize an unbounded body.
- This allowed a malicious or compromised OpenAI-compatible endpoint to trickle bytes or return oversized JSON and defeat the caller's advertised time/size bounds.

### Description
- Updated `RuntimeProvider::chat_completion_non_streaming` to route calls with non-default `config.limits` through the streaming path and `turn_collector::collect_turn`, enforcing `TurnLimits` for both the connect phase and body accumulation; native non-streaming behavior is preserved when limits are unbounded (file: `crates/provider/src/runtime_provider.rs`).
- Added a `NativeNonStreaming` test driver fixture and three regression tests that cover delayed headers (connect-phase timeout), stalled/trickled bodies (body-phase timeout), and responses exceeding `max_response_bytes`, ensuring bounded non-streaming calls use the protected collector rather than the native driver path (file: `crates/provider/src/runtime_provider.rs`, tests under the `#[cfg(test)]` module).

### Testing
- Ran `cargo fmt --check` (succeeded).
- Ran targeted tests `cargo test -p everruns-provider runtime_provider::tests::bounded_non_streaming_calls -- --nocapture` and the provider crate test suite with local-network proxies disabled (`NO_PROXY=localhost,127.0.0.1 cargo test -p everruns-provider`), which passed the new regression tests and the provider crate tests covering streaming/collection behavior.
- Ran `cargo clippy -p everruns-provider --all-targets -- -D warnings` (succeeded).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6aad8c10f964832bbfdf94bd4ed50846)


chaliy:
Rebased onto `main` (clean) and pushed. **CI is now fully green** — the previous red `Build Check` was not this PR's.

## The red was a label-event artifact

`39fcc32` had five CI runs at the same second. One passed; one failed with:

```
run_ci=false: this run executed no jobs, so it has nothing to attest
```

Adding the `codex` and `aardvark` labels fires `pull_request: labeled`, the `ci_event` gate resolves `run_ci=false` for anything outside the `ci:skip-*` set, every job skips, and `Build Check` then hard-fails by design (EVE-939). That leaves a permanent red `Build Check` check-run beside the green one on the same SHA. Rebasing onto a fresh SHA cleared it; the new run is green end to end.

Worth a separate look, though: any PR that gets a non-`ci:skip` label after CI has passed picks up a red required check that only a new push removes. The concurrency split already stops the label run cancelling the real one, which was the actual EVE-939 failure, so the `exit 1` may now be doing more harm than work. I have not touched it here — it is a security guard and not this PR's business. Happy to file it.

## Review

The change is correct. The `ChatDriver` trait already documents the default `chat_completion_non_streaming` as "collects the stream", so routing bounded calls through the collector is the documented fallback applied when limits demand it, not a new behaviour.

Checked for the regression I expected and did not find it: the streaming request sets `stream_options: {include_usage: true}` (`openai_protocol.rs:703`), so bounded calls keep their token usage. `into_response()` carries text, reasoning, tool_calls and metadata across, so nothing else is dropped either.

The three tests are the right ones — asserting `!native_called` proves the native path is genuinely bypassed rather than merely also bounded.

**One note, non-blocking.** `RuntimeProvider::chat_completion_stream` already wraps its result in `limit_stream(limits.after(spent))`, having charged the connect round trip against the budget. The new code then hands that stream to `collect_turn(stream, &config.limits, …)` — the *full* budget again, from a later start. So the limits are applied twice, and the outer copy is strictly looser than the inner one, which is the only reason it is harmless. Compare the trait default at `driver_registry.rs:320`, which passes `&limits.after(spent)` for exactly this reason. Passing the already-charged remainder here too would make the intent obvious to the next reader.

## Validation

`cargo test -p everruns-provider --lib` 323 passed, including all three new `bounded_non_streaming_*` tests; `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

---

chaliy:
Brought `main` in (the branch was 16 commits behind). Merged rather than rebased — a merge commit keeps any existing checkout of this branch valid. No conflicts.

Re-validated on the merged head:

- `cargo clippy -p everruns-provider --all-targets -- -D warnings` — clean
- `NO_PROXY=localhost,127.0.0.1 cargo test -p everruns-provider --lib runtime_provider` — 14 passed, 0 failed, including all three regression tests: `bounded_non_streaming_calls_enforce_the_total_timeout_before_headers`, `..._during_the_body`, and `..._reject_oversized_responses`

Reviewed the change itself as well. Routing bounded calls through `chat_completion_stream` + `collect_turn` is the same path the non-native default already takes (`driver_registry.rs`), so this makes the native opt-in consistent with the fallback rather than inventing a second policy. `ProviderBoundDriver::chat_completion_non_streaming` delegates to `RuntimeProvider`'s method, so the boxed-driver path picks the fix up too. The limits are applied twice — once by `limit_stream` inside `chat_completion_stream` with `limits.after(spent)`, once by the `collect_turn` call here with the full `config.limits` — but the inner bound is the tighter one, so it always fires first and the outer is redundant rather than wrong.

---

PR #3733, opened 2026-09-19
feat(facade): AgentBuilder::backends(HostBackends) for caller-owned persistence

shbodya:
## Summary
- `AgentBuilder::backends(HostBackends)`: let an embedding host hand the facade its own backends bundle (event log, session catalog, checkpoints, …) instead of the in-memory defaults or the `local` profile. The host log stays the sole write path and sequence owner; hooks and the loop are unchanged.
- `BuildError::ConflictingBackends` when combined with `local(..)`.
- `Agent` owns a per-agent `OnceCell<EngineBackends>` for custom bundles → the engine-wide in-memory cell keeps serving default agents, and two agents with different bundles stay isolated inside one `Engine`. `engine::initialize_backends` seeds from the caller's bundle.
- `everruns::HostBackends` re-export; `HostBackends: Debug` (shape only).
- `EventCursor::after_snapshot(session, after, snapshot)` so an out-of-crate `EventReader` can return snapshot-pinned `next_cursor`s (the in-crate logs build them through private fields today).

## Motivation
A Kubernetes host (one binary per tenant, agentgateway in front) needs canonical events + the session catalog in per-tenant Postgres (CloudNativePG) so sessions survive pod restarts, while keeping the facade's typed `on_tool_start` gate. `InProcessRuntimeBuilder::backends` already accepts a custom `EventLog`, but the typed lifecycle hooks are facade-only (private hook capability), so the host API path means re-implementing that plumbing. This is the one-method facade hook that closes the gap. We run it as a carried patch on 0.17.26 today (`PgEventLog`/`PgSessionStore` over sqlx pass the `EventLog` contract tests: read-your-writes, unique monotonic sequence under 20 concurrent appends, snapshot-pinned paging, resume from a fresh `Agent`).

## Test plan
- [x] `cargo test -p everruns --features local --test agent_builder` — 10/10 (new: `custom_backends::{events_land_in_the_provided_event_log, a_fresh_engine_resumes_from_the_provided_backends, separate_backends_stay_isolated_within_one_engine, backends_and_local_are_mutually_exclusive}`)
- [x] `cargo test -p everruns --features local --test engine_sessions` — 4/4
- [x] `cargo test -p everruns --features local --lib` — 125/125; `cargo test -p everruns-host --lib events` — 8/8
- [x] `cargo clippy -p everruns -p everruns-host --features local --all-targets -- -D warnings`, `cargo fmt`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XHN2HAhAsVxXiQgxXp5Xah


chaliy:
@shbodya awesome! Can you rebase on top of main?

PR #3739, opened 2026-09-20
feat(core): record the model the provider actually served — EVE-1049

chaliy:
## What changed

`LlmGenerationMetadata` gains `response_model: Option<String>` — the model the provider reported actually serving the request, alongside the `model` that was asked for. The engine fills it from the completion metadata it already receives, and the OpenTelemetry exporter now reports `gen_ai.response.model` from it, falling back to the requested model when a provider echoes none.

`otel.rs`'s test module moved to its own file to make room under the file-size ratchet (a pure move, no test body changed), and four duplicated metadata fixtures in the Braintrust tests collapsed into one helper.

## Why

EVE-1049 work item B. `model` is what was *requested*, which is routinely an alias that resolves at request time — `claude-sonnet-4-5` served by `claude-sonnet-4-5-20250929`, or an OpenRouter route landing on one of several upstreams. The event carried one string, so the distinction was gone before any exporter saw it.

`otel.rs` then made the loss visible: it pushed `meta.model` into **both** `gen_ai.request.model` and `gen_ai.response.model`, so a span asserted the alias had been confirmed as the serving model. That is the same defect the issue describes in the yolop Logfire exporter, present here too — and it is the only record of which weights produced an answer, which is what a regression in output quality has to be correlated against.

Work item A is in the yolop repository and is not touched here.

## Before / After

The new test fails on `main` and passes here — the behaviour change is exactly this:

```
running 2 tests
test observability::otel::tests::response_model_reports_what_the_provider_served ... FAILED

assertion `left == right` failed: the response attribute must carry the served model
  left: Some("claude-sonnet-4-5")
 right: Some("claude-sonnet-4-5-20250929")
```

After:

```
running 2 tests
test observability::otel::tests::response_model_reports_what_the_provider_served ... ok
test observability::otel::tests::response_model_falls_back_to_the_request_when_unreported ... ok
```

On the wire, an unreported serving model is **absent**, not null, so a downstream exporter can tell "not reported" from "reported as nothing":

```jsonc
// provider echoed no model
{"model": "test-model", "provider": "anthropic", "success": true}
// provider echoed one
{"model": "test-model", "response_model": "test-model-20250929", "provider": "anthropic", "success": true}
```

## Risk

- **Low**
- Additive optional field; `skip_serializing_if` keeps existing payloads byte-identical when unset. The only behaviour change is `gen_ai.response.model`, which becomes *more* accurate and still always present via the fallback.
- The `otel.rs` split is the real review surface. It keeps `#[cfg(test)]` and uses `#[path]`, so the module stays a child of `otel` and `super` resolves unchanged. Test count is identical before and after: host lib 174 both ways.

## Security

- Threat categories reviewed: **TM-OBS**. The new field carries a provider-supplied model identifier into spans and the event log — the same class of value `model` and `provider` already carry, with no credential, user content, or PII exposure. No auth, persistence-schema, or API-surface change.

## One thing to push back on if you disagree

`reason.rs` is also pinned at its length on `main`, and the change there is net +1 line. I paid for it by deleting two comments:

```
- // Extract retry info from completion metadata (if retries occurred)
- // Build LlmGenerationData with retry and compaction info
```

Both restate the statement immediately beneath them, which is what the repo's comment norm says not to write — but I am removing them because a ratchet made me, and that is a worse reason than "they were noise". Flagging it rather than burying it in the diff. If you want either back, the honest alternative is splitting `reason.rs`, which is 2885 lines of execution logic and has no business being touched by this PR.

The `served_model` local is not a line-count dodge: `cargo fmt` wraps the inline form across five lines because it exceeds 100 characters, so the binding is what the formatter wanted anyway.

## Follow-ups

- **Work item A (yolop repo)** — forward `actual_cost_usd`/`effective_cost_usd` and give `llm.generation` spans real duration. Out of scope here: different repository, and this PR is its upstream prerequisite.
- **Other exporters still duplicate the model.** Only `otel.rs` is corrected here. Braintrust and OpenInference were not audited for the same pattern; deliberately left alone rather than changed without a test to justify each.

Resolves [EVE-1049](https://linear.app/everruns/issue/EVE-1049/logfire-exporter-forward-full-llm-usagecost-metadata-and-give-llm) (work item B).

https://claude.ai/code/session_01LpWRWJ1SJBCsvRhCacdve2

---


chaliy:
`Build Check` went red on this PR without a single line of it changing. The failure is not this PR's, and the diff is not involved.

## What happened

`warp-factories[bot]` added the `factory:foreman` label at `16:06:23Z`. Two seconds later a `pull_request` run fired for the `labeled` action on the same head SHA. `Resolve CI event gate` saw a label outside the `ci:skip-*` set and set `run_ci=false`, so every job skipped, and `Verify all jobs passed` exited 1 with:

```
run_ci=false: this run executed no jobs, so it has nothing to attest
```

That refusal is deliberate and correct — it is the EVE-939 guard that stops an all-`skipped` result set reading as a green required check. Nothing here is a bug in the guard's intent.

## The gap

The guard's own comment says "the real run's own Build Check is the one that carries the verdict." That holds only while the real run is the **most recent** `Build Check` on the head SHA. GitHub resolves a required check by the latest check-run of that name, so a label applied *after* CI has finished overwrites a green verdict with a refusal-to-attest — and the PR goes from `clean` to `blocked` with no code change.

The four aardvark PRs in this batch were labelled *before* their CI finished, so their code runs landed last and they are unaffected. This PR was labelled a day later, which is why it is the only one blocked.

## What I did

Re-ran the original code run (`35500476723`, green on this exact SHA yesterday) so its `Build Check` is once more the newest on `ebc1a9c`. No empty commit, no force-push, no code change — re-running the real run is exactly what the guard says should carry the verdict.

Filed [EVE-1077](https://linear.app/everruns/issue/EVE-1077) for the ordering gap itself, since this will recur on any PR that gets labelled after CI completes and the fix must not regress EVE-939.


---

chaliy:
Correction to my comment above: the re-run **did not** unblock this PR. The diagnosis holds, the remedy did not.

Re-running the original code run made a successful `Build Check` the newest check-**run** on `ebc1a9c` (`success`, 16:19:46Z). The PR is still `blocked`, because the label run's failed check-**suite** (`96447652895`) remains on the SHA regardless. Superseding the named check is not enough.

Confirmed against the batch: #3728 and #3731 are both `clean` and both have zero failed suites on their head SHAs; this PR has exactly one and is the only blocked PR. Only a new head SHA clears it.

## What that leaves

Nothing on this branch is wrong. CI passed on this exact commit twice — yesterday, and again just now on the re-run. The block is an artifact of a label applied a day after CI finished.

I am deliberately **not** pushing a no-op commit to clear it. That would work, and it is precisely the "empty commit to kick CI" move that makes a PR's history lie about why it changed.

So this needs one of:

- **any real push to this branch** — the block evaporates with the new SHA; if follow-up review lands here, it clears itself as a side effect; or
- **the [EVE-1077](https://linear.app/everruns/issue/EVE-1077) fix on `main`**, which stops it recurring but will not retroactively clear this SHA; or
- **an admin merge**, since the failing check attests to nothing and the real verdict is green.

I have updated EVE-1077 with the suite-vs-run distinction, and corrected its fix direction: gating `build-check` on `run_ci == 'true'` would *reintroduce* EVE-939, because GitHub counts a skipped required check as passing. The real distinction is not "did this run execute jobs" but "has a real run already reported on this SHA".


---

